PR #19, opened 2019-08-11
PEP 561 compliance

pgjones:
This should allow the typing included with aioquic to be picked up by
mypy and used in projects that use aioquic.


pgjones:
I don't think the build failure is related, is it?

jlaine:
Thanks for the PR, I'm all for it. The CI failure seems to be fallout from a sloppy commit of mine, will fix it

PR #22, opened 2019-08-21
Fix WebSocket example

pgjones:
The connection request is missing a version header (this header is
required).


pgjones:
Note [Hypercorn](https://gitlab.com/pgjones/hypercorn/commit/ad39bb2ad7c5d662b2d5e76e6fb399e7b5cc540f) now supports WebSockets over HTTP/3 if you want to test against something.

jlaine:
Thanks for the patch! I'll try giving hypercorn with websockets a spin using the demo app

PR #34, opened 2019-09-24
[config] add QuicConfiguration.load_cert_chain method

jlaine:


aioquic/quic/configuration.py
```diff
@@ -82,3 +82,20 @@ class QuicConfiguration:
             QuicProtocolVersion.DRAFT_22,
         ]
     )
+
+    def load_cert_chain(
+        self,
+        certfile: str,
```
pgjones: @jlaine What about `PathLike` rather than `str`?

PR #44, opened 2019-11-07
H3 control frames handling and add h3 events

hansroh:
I am making a http3 server with aioquic.

I have some suggestions about control frames and related events especially GOAWAY, DUPLICATE_PUSH and CANCEL_PUSH. These are need to be controlled by server I think.
 
I added 4 methods to h3connection:

- send_cancel_push
- send_duplicate_push
- get_push_id
- close_connection

And add 3 related events:

- DuplicatePushReceived
- PushCanceled
- ConnectionShutdownInitiated

Anyway thanks for your good work.

### DIFF
--- a/examples/interop.py
+++ b/examples/interop.py
@@ -24,7 +24,6 @@ from aioquic.h3.connection import H3_ALPN, H3Connection
 from aioquic.h3.events import DataReceived, HeadersReceived
 from aioquic.quic.configuration import QuicConfiguration
 from aioquic.quic.logger import QuicLogger
-from aioquic.quic.packet import QuicProtocolVersion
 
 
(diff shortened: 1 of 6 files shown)


jlaine:
Thanks for this PR! I'm somewhat swamped at the moment, but I do plan to circle back to it.

Could you possibly split out the GOAWAY support to a separate PR to keep this one focused on PUSH support? Also, Cancelled needs to "L"s

hansroh:
OK, I will try.

jlaine:
> PS: As I know, 'Canceled' is common in US, but if you want, I'll change it.

Ah. In that case go with the US spelling, my mistake.

PR #45, opened 2019-11-07
fix for loop initial declarations error

tianyaqu:
When building with a modern gcc, it fails because of a 'for loop initial declarations error', users have to enable  c99 to overcome that building error,  this PR will fix it.


jlaine:
The issue should be fixed by f1c915d9db6e1e63b02b5fbc58184d70f1baa51f

Issue #47, opened 2019-11-13
Install fails on CentOS7


mikehwang:
Here is the output when installing the aioquic project locally on a CentOS7 machine:

```
# pip3 install -e .                                                                                                                                                                                                
WARNING: Running pip install with root privileges is generally not a good idea. Try `pip3 install --user` instead.                                                                                                                            
Obtaining file:///home/aioquic                                                                                                                                                                                                                
Requirement already satisfied: cryptography>=2.5 in /usr/local/lib64/python3.6/site-packages (from aioquic==0.8.3)                                                                                                                            
Requirement already satisfied: pylsqpack<0.4.0,>=0.3.3 in /usr/local/lib64/python3.6/site-packages (from aioquic==0.8.3)                                                                                                                      
Requirement already satisfied: dataclasses in /usr/local/lib/python3.6/site-packages (from aioquic==0.8.3)                                                                                                                                    
Requirement already satisfied: six>=1.4.1 in /usr/local/lib/python3.6/site-packages (from cryptography>=2.5->aioquic==0.8.3)                                                                                                                  
Requirement already satisfied: cffi!=1.11.3,>=1.8 in /usr/local/lib64/python3.6/site-packages (from cryptography>=2.5->aioquic==0.8.3)                                                                                                        
Requirement already satisfied: pycparser in /usr/local/lib/python3.6/site-packages (from cffi!=1.11.3,>=1.8->cryptography>=2.5->aioquic==0.8.3)                                                                                               
Installing collected packages: aioquic                                                                                                                                                                                                        
  Running setup.py develop for aioquic                                                                                                                                                                                                        
    Complete output from command /usr/bin/python3 -c "import setuptools, tokenize;__file__='/home/aioquic/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'ex
ec'))" develop --no-deps:                                                                                                                                                                                                                     
    running develop                                                                                                                                                                                                                           
    running egg_info                                                                                                                                                                                                                          
    writing src/aioquic.egg-info/PKG-INFO                                                                                                                                                                                                     
    writing dependency_links to src/aioquic.egg-info/dependency_links.txt                                                                                                                                                                     
    writing requirements to src/aioquic.egg-info/requires.txt                                                                                                                                                                                 
    writing top-level names to src/aioquic.egg-info/top_level.txt                                                                                                                                                                             
[...]

PR #48, opened 2019-11-14
Fix CentOS7 build issue

mikehwang:
https://github.com/aiortc/aioquic/issues/47

Tested with CentOS7 and Ubuntu 18.04


jlaine:
I'm kind of torn between the approach in #45 and this PR.

If we go with this PR, do these flags work with compilers other than gcc and clang?

PR #50, opened 2019-11-19
sending GOAWAY frame and receiving ConnectionShutdownInitiated event

hansroh:
add,

- h3.H3Connection.connection_close (last_stream_id) for a server
- h3.events.ConnectionShutdownInitiated for a client

Also I closed my previous PR. 
After this PR is confirmed, I'll make PUSH support PR I think.

I gladly wait for your opinions.

### DIFF
--- a/src/aioquic/h3/connection.py
+++ b/src/aioquic/h3/connection.py
@@ -17,7 +17,6 @@ from aioquic.h3.exceptions import NoAvailablePushIDError
 from aioquic.quic.connection import QuicConnection, stream_is_unidirectional
 from aioquic.quic.events import QuicEvent, StreamDataReceived
 from aioquic.quic.logger import QuicLoggerTrace
-from aioquic.quic.packet import QuicErrorCode
 
 logger = logging.getLogger("http3")
 
--- a/src/aioquic/h3/connection.py
+++ b/src/aioquic/h3/connection.py
@@ -17,6 +17,7 @@ from aioquic.h3.exceptions import NoAvailablePushIDError
 from aioquic.quic.connection import QuicConnection, stream_is_unidirectional
 from aioquic.quic.events import QuicEvent, StreamDataReceived
 from aioquic.quic.logger import QuicLoggerTrace
+from aioquic.quic.packet import QuicErrorCode
 
 logger = logging.getLogger("http3")
 
@@ -129,7 +130,7 @@ def encode_settings(settings: Dict[int, int]) -> bytes:
     return buf.data
 
 
-def parse_id(data: bytes) -> int:
+def parse_uint_var(data: bytes) -> int:
     buf = Buffer(data=data)
     _id = buf.pull_uint_var()
     assert buf.eof()
@@ -225,7 +226,7 @@ class H3Connection:
         self._stream: Dict[int, H3Stream] = {}
 
         self._max_push_id: Optional[int] = 8 if self._is_client else None
-        self._max_client_init_bi_stream_id: int = 0 if not self._is_client else None
+        self._max_client_bidi_stream_id: int = 0 if not self._is_client else None
         self._next_push_id: int = 0
 
         self._local_control_stream_id: Optional[int] = None
@@ -362,19 +363,19 @@ class H3Connection:
             stream_id, encode_frame(FrameType.HEADERS, frame_data), end_stream
         )
 
-    def close_connection(self, last_stream_id: int = None) -> None:
+    def shutdown(self, last_stream_id: int = None) -> None:
         """
-        Close a connection, emitting a GOAWAY frame.
-        :param last_stream_id: Client-initailized stream ID that can be responsed before connection is closed.
+        Send a GOAWAY frame to client from server.
+        :param last_stream_id: Client-initiated stream ID that can be served before connection is closed.
         """
         # client need not send GOAWAY frame
         assert not self._is_client, "Client must not send a goaway frame"
         if last_stream_id is None:
-            last_stream_id = self._max_client_init_bi_stream_id
+            last_stream_id = self._max_client_bidi_stream_id
         else:
             assert (
-                last_stream_id <= self._max_client_init_bi_stream_id
-            ), "Unknown stream ID"
+                last_stream_id <= self._max_client_bidi_stream_id
+            ), "Unissued stream ID"
         self._quic.send_stream_data(
             self._local_control_stream_id,
             encode_frame(FrameType.GOAWAY, encode_uint_var(last_stream_id)),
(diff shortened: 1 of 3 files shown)


src/aioquic/h3/connection.py
```diff
@@ -128,11 +129,11 @@ def encode_settings(settings: Dict[int, int]) -> bytes:
     return buf.data
 
 
-def parse_max_push_id(data: bytes) -> int:
+def parse_id(data: bytes) -> int:
```
jlaine: Maybe `parse_uint_var` would be a better name?

src/aioquic/h3/connection.py
```diff
@@ -224,6 +225,7 @@ def __init__(self, quic: QuicConnection):
         self._stream: Dict[int, H3Stream] = {}
 
         self._max_push_id: Optional[int] = 8 if self._is_client else None
+        self._max_client_init_bi_stream_id: int = 0 if not self._is_client else None
```
jlaine: I think I'd prefer `_max_client_bidi_stream_id`

src/aioquic/h3/connection.py
```diff
@@ -360,6 +362,24 @@ def send_headers(
             stream_id, encode_frame(FrameType.HEADERS, frame_data), end_stream
         )
 
+    def close_connection(self, last_stream_id: int = None) -> None:
+        """
+        Close a connection, emitting a GOAWAY frame.
+        :param last_stream_id: Client-initailized stream ID that can be responsed before connection is closed.
```
jlaine: I'm not sure `close_connection` is very descriptive, since the definition of `GOAWAY` explicitly states: "GOAWAY by itself does not close a connection". The RFC uses the term "shutdown", so maybe "shutdown_connection"?

This docstring needs improvement:

- Mention only servers can call this method
- Typo: "initialized" 
- I'm not sure `last_stream_id` is a good name, as this is in fact the first stream which will never get a response. Maybe simply `stream_id` would be better. Also, why do we want this part of the API? Can't we automatically determine the last stream for which we fully served the response?

src/aioquic/h3/connection.py
hansroh: agree

src/aioquic/h3/connection.py
hansroh: ok

src/aioquic/h3/connection.py
hansroh: - 'close_connection' follows H2Connection API, but also I don't like this. I think 'shutdown' is fine. 

- My English is so poor ;-) but I think 'initiated' is more proper than 'initialized' by RFC section-7.2.6)

- 'last_stream_id' is also follow [H2Connection API](https://python-hyper.org/projects/h2/en/stable/api.html#h2.connection.H2Connection.close_connection). And I think it is more explicit and it is good.

   And I can't understand exactly your words. 'last_stream_id' is optional but I think H3Connection should give explicit way for server administrative activity. For example, under exhaust of system memory situation, server need both: 

  - rapid recovering by ignoring some of already received requests
  - but graceful shutdown manner by notifying explicit last stream ID to serve in spite of server emergency

TLTR, 

- I agree to change 'shutdown'
- 'last_stream_id' is better and it is very necessary

jlaine: I had not realized you were mimicking H2Connection's API.. now I have conflicting feelings about naming.

In any case can you please add full test coverage for the new code?

hansroh: I think I already make full test coverage. Anyway, when your master branch became stable, I will merge and retry.

src/aioquic/h3/connection.py
jlaine: This function is used when you have some bytes which contain *only* a variable-length integer, hence the EOF check, to check all bytes have been consumed.

src/aioquic/h3/connection.py
```diff
@@ -362,19 +362,19 @@ def send_headers(
...
-        Close a connection, emitting a GOAWAY frame.
-        :param last_stream_id: Client-initailized stream ID that can be responsed before connection is closed.
+        Send a GOAWAY frame to client from server.
+        :param last_stream_id: Client-initiated stream ID that can be served before connection is closed.
         """
         # client need not send GOAWAY frame
         assert not self._is_client, "Client must not send a goaway frame"
         if last_stream_id is None:
-            last_stream_id = self._max_client_init_bi_stream_id
+            last_stream_id = self._max_client_bidi_stream_id
```
jlaine: What happens if `self._max_client_bidi_stream_id` is None?

src/aioquic/h3/connection.py
```diff
@@ -362,19 +362,19 @@ def send_headers(
...
         else:
             assert (
-                last_stream_id <= self._max_client_init_bi_stream_id
-            ), "Unknown stream ID"
+                last_stream_id <= self._max_client_bidi_stream_id
```
jlaine: Same here, if `self._max_client_bidi_stream_id` is None, this will bomb

src/aioquic/h3/connection.py
hansroh: Maybe we are different concept about client-initiated stream ID.

Client-initiated stream ID means just one of client's request stream IDs. Not push stream ID nor uni-directional stream ID.

H2Connection explain that `last_stream_id` is the last stream which was processed by the sender.

FYI [HTTP3 dfraft](https://tools.ietf.org/html/draft-ietf-quic-http-23#section-5.2)

I'm not very good at English. Any suggestion?

hansroh: Nerver mind about my PRs. I have my own solution for now.

jlaine:
This isn't very friendly, would you mind detailing your solution?

hansroh:
Sorry, It just mean I'll catch up your code. I can't wait my PRs without deadline because my coding schedule. So I decide I made temporary my own APIs, and update after your APIs eventually.

I made sub-class of H3Connection and new events on my codes:

- https://gitlab.com/hansroh/aquests/blob/master/aquests/protocols/http3/connection.py
- https://gitlab.com/hansroh/aquests/blob/master/aquests/protocols/http3/events.py

As you change your code, I will update my code and I hope my codes will be removed completely at some moment.

FYI, my server implementation:

- https://gitlab.com/hansroh/skitai/blob/master/skitai/backbone/http3_server.py
- https://gitlab.com/hansroh/skitai/blob/master/skitai/handlers/http3_handler.py

jlaine:
This is not very satisfactory, if you did not intend to put in the work to get the PR into a mergeable shape (including full unit tests like the rest of aioquic) it wasn't very considerate wasting my limited time.

Do come back with a PR when you actually want it to have a chance to be merged.

hansroh:
You maybe misunderstand. I don't care about merging my PRs. I do care my project working properly with your project and I found solutions for now.

 I suggested some ideas and it's your call to use or not.

For your clearance, I have no interest any more to merge.

Good luck to you and your project.

PR #58, opened 2020-01-08
Add local port for the client

keepsimple1:
I am trying to run a QUIC client bind with a certain port and found there is no such option today. Hence submitting this PR. Please let me know if it makes sense.  Thank you for your work.

Unit test and manual test are done.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,5 @@
 /build
 /dist
 /docs/_build
+.idea
+venv
--- a/.gitignore
+++ b/.gitignore
@@ -8,5 +8,3 @@
 /build
 /dist
 /docs/_build
-.idea
-venv
--- a/examples/http3_client.py
+++ b/examples/http3_client.py
@@ -399,11 +399,7 @@ if __name__ == "__main__":
         "-v", "--verbose", action="store_true", help="increase logging verbosity"
     )
     parser.add_argument(
-        "-p",
-        "--local-port",
-        type=int,
-        default=0,
-        help="local port to bind for connections",
+        "-p", "--local-port", type=int, default=0, help="local port to bind for connections",
     )
 
     args = parser.parse_args()
--- a/examples/http3_client.py
+++ b/examples/http3_client.py
@@ -399,7 +399,7 @@ if __name__ == "__main__":
         "-v", "--verbose", action="store_true", help="increase logging verbosity"
     )
     parser.add_argument(
-        "-p",
+        "-lp",
         "--local-port",
         type=int,
         default=0,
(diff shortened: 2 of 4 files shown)


src/aioquic/asyncio/client.py
```diff
@@ -21,6 +21,7 @@
     create_protocol: Optional[Callable] = QuicConnectionProtocol,
     session_ticket_handler: Optional[SessionTicketHandler] = None,
     stream_handler: Optional[QuicStreamHandler] = None,
+    local_port: Optional[int] = None,
```
jlaine: The variable name is fine, could we have a docstring for it please?

.gitignore
```diff
@@ -8,3 +8,5 @@
 /build
 /dist
 /docs/_build
+.idea
```
jlaine: Maybe keep this out of the PR as it's not related

src/aioquic/asyncio/client.py
jlaine: I don't think making it `Optional` is right, couldn't we initialize it straight to `0`?

tests/test_asyncio.py
```diff
@@ -287,6 +287,11 @@ def test_connect_timeout(self):
                 )
             )
 
+    def test_connect_local_port(self):
```
jlaine: Thank you taking the time to write a test, it makes we want to get this merged :)

jlaine:
Do you still plan to complete this PR with the requested changes?

jlaine:
Could you please remind me what the motivation for this PR is?

src/aioquic/asyncio/client.py
keepsimple1: done.

.gitignore
keepsimple1: done.

keepsimple1:
I apologize about my lengthy delay.  I've updated my diff.  My motivation is that: I was using the client to test a NAT-traversal case, where I need the client to let the server know (via signaling) its port before setting up the QUIC connection.  In such cases, it is quite useful to let client to bind to a particular port.

examples/http3_client.py
```diff
@@ -396,6 +398,13 @@ def save_session_ticket(ticket: SessionTicket) -> None:
     parser.add_argument(
         "-v", "--verbose", action="store_true", help="increase logging verbosity"
     )
+    parser.add_argument(
+        "-lp",
```
jlaine, line 402: I'm not sure the short version gains us anything, maybe drop it?

keepsimple1: I've dropped it. Personally I liked it, but it's not a must-have, and can be added later if needed.

jlaine:
Could you fix the linter error please?

keepsimple1:
Sure, the linter error is fixed now.

keepsimple1:
Thank you for merging!

PR #62, opened 2020-02-06
[recovery] split out congestion control

jlaine:


src/aioquic/quic/recovery.py
```diff
@@ -34,6 +34,69 @@ def __init__(self) -> None:
...
+
+    def on_packets_expired(self, packets: Iterable[QuicSentPacket]) -> None:
+        for packet in packets:
+            self.bytes_in_flight -= packet.sent_bytes
+
+    def on_packets_lost(self, packets: Iterable[QuicSentPacket], now: float) -> None:
+        lost_largest_time = 0.0
+        for packet in packets:
+            self.bytes_in_flight -= packet.sent_bytes
+            lost_largest_time = packet.sent_time
```
sethmlarson: Is `last_largest_time` supposed to be the maximum value of loss?

jlaine: `lost_largest_time` refers to the most recent `sent_time` of the lost packets. I should probably make a note to the effect that `packets` are sorted by sent time. Does that make sense?

sethmlarson: Yeah! Totally misread, disregard me :)

PR #82, opened 2020-04-09
Initial attempt at a DNS over QUIC client/server

tigertoes:
One thing to note about this pull request is that it depends on dnslib, which has *not* been included into `requirements/doc.txt` as this is an example piece of code. I'm happy to consider an alternative (including spinning it off into another repository) if this isn't the right thing to do.

### DIFF
--- a/examples/doq_client.py
+++ b/examples/doq_client.py
@@ -1,0 +1,154 @@
+import argparse
+import asyncio
+import json
+import logging
+import pickle
+import ssl
+from typing import Optional, cast
+
+from dnslib.dns import QTYPE, DNSQuestion, DNSRecord
+
+from aioquic.asyncio.client import connect
+from aioquic.asyncio.protocol import QuicConnectionProtocol
+from aioquic.quic.configuration import QuicConfiguration
+from aioquic.quic.events import QuicEvent, StreamDataReceived
+from aioquic.quic.logger import QuicLogger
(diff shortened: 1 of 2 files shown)


jlaine:
Welcome back @tigertoes, thank you for this cool stuff!

One question comes to mind: on the server side, the request to the upstream resolver seems to be synchronous, is that right?

tigertoes:
You're right, it's a blocking and only done to get something working, it's not necessarily the right answer. Happy to change it if you think it's more appropriate.

jlaine:
There is an async DNS package here: https://github.com/saghul/aiodns

Does it cover your needs? Failing that we could probably just run the lookup code in an executor.

jlaine:
On the server side, as your code is nicely structured I'm pondering merging it straight into the HTTP/3 server example, that way we could have a public-facing server at quic.aiortc.org

tigertoes:
I'm not sure merging it into the HTTP/3 server is the right call - DoQ aims to have not only a separate ALPN, but run on a completely different port. Unless of course, you were intending the HTTP/3 example server to listen on both 443 and 874?

I'll take a look at making the upstream resolution happen asynchronously.

jlaine:
Ah I hadn't realised it's supposed to run on a separate port! Let's keep things separate in that case.

PR #85, opened 2020-04-16
add peer_address public method

keepsimple1:
This is to provide a public method `peer_address` for `QuicConnection`, `H0Connection`, `H3Connection` and use it in the `http3_server.py` example.

### DIFF
--- a/examples/http3_server.py
+++ b/examples/http3_server.py
@@ -257,8 +257,7 @@ class HttpServerProtocol(QuicConnectionProtocol):
                 path_bytes, query_string = raw_path, b""
             path = path_bytes.decode()
 
-            # FIXME: add a public API to retrieve peer address
-            client_addr = self._http._quic._network_paths[0].addr
+            client_addr = self._http.peer_address()
             client = (client_addr[0], client_addr[1])
 
             handler: Handler
--- a/src/aioquic/h0/connection.py
+++ b/src/aioquic/h0/connection.py
@@ -1,7 +1,7 @@
-from typing import Dict, List
+from typing import Dict, List, Optional
 
 from aioquic.h3.events import DataReceived, H3Event, Headers, HeadersReceived
-from aioquic.quic.connection import QuicConnection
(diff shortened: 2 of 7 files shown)


keepsimple1:
There is a test failure in windows-latest 3.8.  It seems not related to my diff, and it seems happened in some [previous commits](https://github.com/aiortc/aioquic/runs/588833423) as well. Is this a known issue?  @jlaine 

pasted from the log:
```
FAIL: test_send_max_stream_data_retransmit (tests.test_connection.QuicConnectionTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "d:\a\aioquic\aioquic\tests\test_connection.py", line 1591, in test_send_max_stream_data_retransmit
    self.assertEqual(stream.max_stream_data_local_sent, 2097152)
AssertionError: 0 != 2097152
```

jlaine:
Thanks for taking the time to put this PR together, and double thanks for writing unit tests!

There are a couple of points I'm not entirely comfortable with:

- so far, the QUIC stack has been as network-agnostic as possible. This means there is no guarantee about the form of the "peer address": it could be a traditional (address, port) tuple, but it doesn't have to be

- the "peer_address" API obscures the fact there could be multiple network paths (although admitedly aioquic only uses one at any given time)

Any thoughts on this?

keepsimple1:
Not a problem.  Here's my thoughts: 

1. In the patch, `peer_address()` returns `NetworkAddress` (if not None), which has the type of `Any`, i.e. does not constrain it to any particular address format. I agree that my comments seem to limit to only IP endpoint tuple. I can update the comments section to clarify this.

2. It's my understanding that the current / active peer address is always the 1st in `_network_paths`, and if changes, it will promote to `_network_paths[0]` .  If that's true, I can add more comments to clarify that.  If not true, I probably have to change code to handle different cases.

Thanks.

keepsimple1:
I've updated the diff based on my previous comments.

PR #92, opened 2020-05-02
Properly calculate Session Ticket's obfuscated age

rmarx:
Per https://tools.ietf.org/html/rfc8446#section-4.2.11.1:
> Note that the "ticket_lifetime" field in the NewSessionTicket message
   is in seconds but the "obfuscated_ticket_age" is in milliseconds.

This caused problems in interop with QUIC stacks, as the valid time of a ticket was suddenly off by a factor of 1000 (e.g., leading to a ticket only being usable for 10 seconds instead of a few hours/days). 

NOTE: I'm not sure I'm following the logic entirely, but it seems the server doesn't really check the lifetime of a ticket it previously issued either? _build_session_ticket() seems to just take the current timestamp + lifetime, meaning it's always valid? But I must be reading it wrong, at 10:45 PM :)

### DIFF
--- a/src/aioquic/tls.py
+++ b/src/aioquic/tls.py
@@ -1161,7 +1161,7 @@ class SessionTicket:
 
     @property
     def obfuscated_age(self) -> int:
-        age = int((utcnow() - self.not_valid_before).total_seconds())
+        age = int((utcnow() - self.not_valid_before).total_seconds() * 1000)
         return (age + self.age_add) % (1 << 32)
 
 


jlaine:
Hi Robin!

Thanks for looking into this, I need to re-read the code to check whether `obfuscated_age` is being calculated correctly.

I believe the ticket age validation is done correctly however as it relies on `ticket_lifetime`.

Jeremy

jlaine:
Update: if you rebase on top of master, all tests should pass

rmarx:
So... I was kind of lazy and just opened this PR by editing the file in github directly (I don't have a full checkout of aioquic master at this time), so not really sure if I can do a rebase inside github?

jlaine:
I have rebased your PR, github had created a "patch-1" branch for you. How can I check this change, do you have a peer against which you had an error?

rmarx:
Yes, the error was highly reproducible against the picoquic public endpoint (but also against most others, though they seems to have a higher tolerance). I did a normal 1-RTT connection and stored the ticket. Then waited 10s+ to use it for a 0-RTT connection and it never succeeded. With this patch, it now always succeeds (succeed = tls.early_data_accepted is True).

And quite frankly, SIR, I'm appalled that you don't just believe me at face value, PFOE! ;)

PR #109, opened 2020-06-04
WIP: [connection] Add Reset Stream API

Aniketh01:


Aniketh01:
Hi @jlaine, 

Thank you for your guidance on this patch.

Meanwhile, would you mind taking a look at how the "final_size" is created? I'm not quite sure if that is the proper way..
 
This is a work in progress patch. I think the current PR creates a signal that can be used to denote whether a stream is a Reset Stream or not. I think we need to add code in which, we deal with the question, "What should we do after a Reset stream is received?".

Do you have any suggestion on how to move forward?

src/aioquic/quic/connection.py
```diff
@@ -289,6 +291,7 @@ def __init__(
         self._state = QuicConnectionState.FIRSTFLIGHT
         self._stateless_retry_count = 0
         self._streams: Dict[int, QuicStream] = {}
+        self._streamFrames: Dict[int, QuicStreamFrame] = {}
```
jlaine, line 294: This seems to be unused?

src/aioquic/quic/connection.py
```diff
@@ -233,6 +234,7 @@ def __init__(
         self._ack_delay = K_GRANULARITY
         self._close_at: Optional[float] = None
         self._close_event: Optional[events.ConnectionTerminated] = None
+        self._reset_event: Optional[events.StreamReset] = None
```
jlaine, line 237: This doesn't seem to be used? Also, let's try and keep things sorted alphabetically :)

src/aioquic/quic/connection.py
```diff
@@ -85,6 +85,7 @@
 RETIRE_CONNECTION_ID_CAPACITY = 1 + 8
 STREAMS_BLOCKED_CAPACITY = 1 + 8
 TRANSPORT_CLOSE_FRAME_CAPACITY = 1 + 8 + 8 + 8  # + reason length
+RESET_STREAM_CAPACITY = 1 + 8 + 8
```
jlaine, line 88: There are 3 varints in a RESET_STREAM frame, so you need another +8

src/aioquic/quic/connection.py
```diff
@@ -2317,6 +2346,17 @@ def _write_application(
                         ),
                     )
 
+            # RESET STREAM
+            for stream in self._streams.values():
+                if self._reset_pending:
```
jlaine, line 2351: A reset is specific to a stream, so the _reset_pending flag should an attribute of the stream, not the connection

src/aioquic/quic/connection.py
```diff
@@ -2718,3 +2758,33 @@ def _write_streams_blocked_frame(
...
+        buf = builder.start_frame(
+            frame_type=frame_type,
+            capacity= RESET_STREAM_CAPACITY,
+            handler=self._on_reset_stream_delivery,
+        )
+        buf.push_uint_var(stream.stream_id)
+        buf.push_uint_var(error_code)
+
+        length = len(size_uint_var(stream.stream_id))
+        final_size = size_uint_var(stream.stream_id) + length
```
jlaine, line 2779: I don't think this is correct, I believe final_size represents how much data has been sent on the stream

Aniketh01: Is `_send_highest` attribute from QuicStream which denotes the final size?

src/aioquic/quic/connection.py
```diff
@@ -398,6 +400,27 @@ def change_connection_id(self) -> None:
...
+        Reset a stream.
+
+        :param stream_id: The ID of the stream to reset.
+        :param error_code: (optional) An error code indicating why the connection is
+                           being reset. Default: `QuicErrorCode.NO_ERROR`
+        """
+        stream = self._streams[stream_id] = QuicStream(stream_id=stream_id)
+
+        if self._state not in END_STATES:
+            self._reset_event = events.StreamReset(
```
jlaine, line 418: Same comment as for `_reset_pending` : the error code is specific to a stream. I think it would be sufficient to store `_reset_error_code` on the stream.

src/aioquic/quic/connection.py
```diff
@@ -398,6 +400,27 @@ def change_connection_id(self) -> None:
...
+        error_code: int = QuicErrorCode.NO_ERROR,
+    ) -> None:
+        """
+        Reset a stream.
+
+        :param stream_id: The ID of the stream to reset.
+        :param error_code: (optional) An error code indicating why the connection is
+                           being reset. Default: `QuicErrorCode.NO_ERROR`
+        """
+        stream = self._streams[stream_id] = QuicStream(stream_id=stream_id)
```
jlaine, line 415: I don't think this is right, we don't want to *create* a new stream. We want to reset an existing stream, and asking for a reset on a stream which doesn't exist should raise an exception.

jlaine:
Thanks for all the work, which served as the basis for #110

PR #184, opened 2021-04-12
Only override configuration.server_name if it's not already set

silverbucket:
This PR preserves the value of `configuration.server_name` if it has already been set. 

We use an SNI that's different from the hostname, so we need to set it manually. As the code is now, we're unable to do so because `server_name` is always overwritten with the hostname. This patch fixes that issue.


jlaine:
Thanks for this!

PR #192, opened 2021-05-19
Change TLS behavior

possessedbysatan:
Hi! I just noticed that there is no support for Ed25519 (hash signature). 

`ED25519` is indeed declared in the `SignatureAlgorithm` enum class (along with many other signature algorithms), however in `_server_handle_hello`  there's support for only a handful (RSA, SECP256R1).  I do know that browsers and most CA authorities use only a few algorithms according to the CA/Browser forum, which kinda sucks to be honest. However, [OpenSSL supports](https://blog.pinterjann.is/ed25519-certificates.html) Ed25519 hash signatures just fine, so why not add support for it?

Finally, I also noticed that aioquic uses by default `SECP256R1` in _client_send_hello, which seems to happen because `self._supported_groups` is initialized with `Group.SECP256R1`. I've changed the default to X25519, if supported, using SECP256R1 as last resort. I am actually kinda oblivious to the RFCs and I don't know if SECP256R1 should be tried first, either because of the RFC or compatibility reasons. Based on [this](https://datatracker.ietf.org/doc/html/draft-ietf-ipsecme-safecurves-03#section-4) I'd suggest using X25519 by default.

You might want to take a look at this, it's probably not ready to be merged as-is, since I didn't take into account anything regarding compatibility: I just added the missing bits using pyca's ed25519 module. aioquic, however, seems to work just fine.

Signed-off-by: JUAN MÉNDEZ <vpsink@gmail.com>

### DIFF
--- a/src/aioquic/tls.py
+++ b/src/aioquic/tls.py
@@ -33,6 +33,7 @@ from cryptography.hazmat.primitives.asymmetric import (
     rsa,
     x448,
     x25519,
+    ed25519,
 )
 from cryptography.hazmat.primitives.kdf.hkdf import HKDFExpand
 from cryptography.hazmat.primitives.serialization import Encoding, PublicFormat
@@ -1043,6 +1044,7 @@ CIPHER_SUITES = {
 }
 
 SIGNATURE_ALGORITHMS: Dict = {
+    SignatureAlgorithm.ED25519: (None, hashes.SHA512),
     SignatureAlgorithm.ECDSA_SECP256R1_SHA256: (None, hashes.SHA256),
     SignatureAlgorithm.ECDSA_SECP384R1_SHA384: (None, hashes.SHA384),
     SignatureAlgorithm.ECDSA_SECP521R1_SHA512: (None, hashes.SHA512),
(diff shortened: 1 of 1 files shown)


src/aioquic/tls.py
```diff
@@ -1838,12 +1854,20 @@ def _server_handle_hello(
                 )
 
             # send certificate verify
-            signature = self.certificate_private_key.sign(
-                self.key_schedule.certificate_verify_data(
-                    b"TLS 1.3, server CertificateVerify"
-                ),
-                *signature_algorithm_params(signature_algorithm),
-            )
+            if isinstance(self.certificate_private_key, ed25519.Ed25519PrivateKey):
```
jlaine, line 1857: Instead of adding this if / else, couldn't you adjust the `signature_algorithm_params` method?

src/aioquic/tls.py
```diff
@@ -1571,13 +1575,21 @@ def _client_handle_certificate_verify(self, input_buf: Buffer) -> None:
...
         try:
-            self._peer_certificate.public_key().verify(
-                verify.signature,
-                self.key_schedule.certificate_verify_data(
-                    b"TLS 1.3, server CertificateVerify"
-                ),
-                *signature_algorithm_params(verify.algorithm),
-            )
+            if isinstance(self._peer_certificate.public_key(), ed25519.Ed25519PublicKey):
+                self._peer_certificate.public_key().verify(
```
jlaine, line 1579: Same as below, adjusting `signature_algorithm_params` would avoid this repeated code.

jlaine:
I'm going to close this PR as it introduces some unnecessary changes and has no unit tests. An alternative implementation is provided in PR #202

PR #194, opened 2021-05-28
fix: H3 application ConnectionClose code value should be 256 and not 0

vranem1:
It looks like h3 client is sending "Application ConnectionClose" with wire_code 0.
But https://tools.ietf.org/html/draft-ietf-quic-http-34#section-8.1
specifies that H3_NO_ERROR is 256, not 0.


jlaine:
This patch won't land, asyncio.client is not just about HTTP/3. The try/except is a last resort option, you probably want to add an explicit `close()` in the HTTP/3 example itself.

vranem1:
@jlaine: like this?

jlaine:
Yes that looks better, could you rebase your PR against main please?

vranem1:
rebased

PR #198, opened 2021-06-14
Fix mutable default arguments in http3_client example

johnthagen:
Fix usage of mutable default arguments.

For an explanation of why mutable default arguments are a bad practice, see https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments.

### DIFF
--- a/examples/http3_client.py
+++ b/examples/http3_client.py
@@ -251,7 +251,7 @@ class HttpClient(QuicConnectionProtocol):
 async def perform_http_request(
     client: HttpClient,
     url: str,
-    data: str,
+    data: Optional[str],
     include: bool,
     output_dir: Optional[str],
 ) -> None:
@@ -349,7 +349,7 @@ def save_session_ticket(ticket: SessionTicket) -> None:
 async def run(
     configuration: QuicConfiguration,
     urls: List[str],
-    data: str,
+    data: Optional[str],
     include: bool,
     output_dir: Optional[str],
     local_port: int,
--- a/examples/http3_client.py
+++ b/examples/http3_client.py
@@ -140,9 +140,6 @@ class HttpClient(QuicConnectionProtocol):
         """
         Perform a GET request.
         """
-        if headers is None:
-            headers = {}
-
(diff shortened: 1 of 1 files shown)


jlaine:
Hi! This PR makes sense, but could you please fix the linter errors and code coverage drop?

jlaine:
This PR makes sense, but it looks as though I need to fix a linter error unrelated to your changes.

johnthagen:
@jlaine Sounds great. Feel free to merge `main` into this branch when you are done as needed, or ping me to do it.

examples/http3_client.py
```diff
@@ -129,26 +136,39 @@ def __init__(self, *args, **kwargs) -> None:
...
         )
 
-    async def post(self, url: str, data: bytes, headers: Dict = {}) -> Deque[H3Event]:
+    async def post(
+        self, url: str, data: bytes, headers: Optional[Dict] = None
+    ) -> Deque[H3Event]:
         """
         Perform a POST request.
         """
+        if headers is None:
```
jlaine, line 156: This seems unnecessary if HttpRequest accepts None

examples/http3_client.py
```diff
@@ -129,26 +136,39 @@ def __init__(self, *args, **kwargs) -> None:
...
         )
 
-    async def websocket(self, url: str, subprotocols: List[str] = []) -> WebSocket:
+    async def websocket(
+        self, url: str, subprotocols: Optional[List[str]] = None
+    ) -> WebSocket:
         """
         Open a WebSocket.
         """
+        if subprotocols is None:
```
jlaine, line 169: This doesn't seem necessary, as there is an `if subprotocols` later in this function

examples/http3_client.py
johnthagen: Fixed.

examples/http3_client.py
johnthagen: Fixed.

jlaine:
Thanks for this @johnthagen , merged!

PR #213, opened 2021-07-22
Correct required connect headers

silverbucket:
Based on the IETF spec for both CONNECT and CONNECT-UDP, the following header validation adjustments were made:

* `:scheme` and `:path` MUST be omitted
* `:authority` MUST? be present

https://quicwg.org/base-drafts/draft-ietf-quic-http.html#section-4.2
https://tools.ietf.org/id/draft-schinazi-masque-connect-udp-00.html#name-the-connect-udp-method

Based on these linked documents, while it's clear that `:scheme` and `:path` must be omitted in both cases, it's unclear if `:authority` is actually required or if that's just for HTTP/3. So perhaps some adjustments still need to be made, submitting here for your thoughts. Should `:authority` be validated or not?

Additionally, I was unable to get the examples to work using the documentation in the README, so I did not make any changes there and there's a possibility this PR will break that code.

```bash
$ venv/bin/pip install -e .
... (successful)
$ venv/bin/pip install aiofiles asgiref dnslib httpbin starlette wsproto
... (successful)
$ venv/bin/python examples/http3_client.py --ca-certs tests/pycacert.pem https://localhost:4433/
Traceback (most recent call last):
  File "examples/http3_client.py", line 12, in <module>
    import wsproto
ModuleNotFoundError: No module named 'wsproto'
```

### DIFF
--- a/src/aioquic/h3/connection.py
+++ b/src/aioquic/h3/connection.py
@@ -285,7 +285,7 @@ def validate_request_headers(headers: Headers) -> None:
             # we use it for the WebSocket demo.
             (b":method", b":scheme", b":authority", b":path", b":protocol")
         ),
-        required_pseudo_headers=frozenset((b":method", b":scheme", b":path")),
+        required_pseudo_headers=frozenset((b":method", b":authority")),
     )
 
 
--- a/tests/test_h3.py
+++ b/tests/test_h3.py
@@ -1703,7 +1703,7 @@ class H3ParserTest(TestCase):
             validate_request_headers([(b":method", b"GET")])
         self.assertEqual(
             cm.exception.reason_phrase,
-            "Pseudo-headers [b':path', b':scheme'] are missing",
+            "Pseudo-headers [b':authority'] are missing",
         )
 
         # empty :authority pseudo-header for http/https


jlaine:
Hello Nick! Let's keep the discussion focused on the HTTP/3 headers, a PR is not the right place to discuss unrelated packaging topics.

I don't think `:scheme` and `:path` are forbidden in HTTP/3, there is at least one example of CONNECT over HTTP/3 which requires both these headers: WebTransport. See:

https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3#section-3.3

I'm not sure `CONNECT-UDP` falls within the scope of aioquic, is this actually something that is defined for HTTP/3?

silverbucket:
> I don't think `:scheme` and `:path` are forbidden in HTTP/3, there is at least one example of CONNECT over HTTP/3 which requires both these headers: WebTransport. See:
> 
> https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3#section-3.3

If I understand correctly, t looks like the WebTransport functionality extends upon the HTTP/3 CONNECT method. See the sections 3.1, 3.2 -- technically the WebTransport functionality shouldn't be accepted unless specified in the SETTINGS frame.


> I'm not sure `CONNECT-UDP` falls within the scope of aioquic, is this actually something that is defined for HTTP/3?

It seems so, taken from: https://datatracker.ietf.org/doc/draft-ietf-masque-connect-udp/

_"The CONNECT-UDP method is defined for all versions of HTTP.  UDP
  payloads are sent using HTTP Datagrams [HTTP-DGRAM]."_ 

However, I will also point out that it seems I was referencing an early version of the draft initially, which specifically omits the `:scheme` and `:path` headers. While with the latest draft (July 2021) they explicitly require them. So, similar to WebTransport in that case.

HTTP/3 specifically omits them, while CONNECT-UDP and WebTransport specifically add them back in.


> Hello Nick! Let's keep the discussion focused on the HTTP/3 headers, a PR is not the right place to discuss unrelated packaging topics.

Agreed, just wanted to make sure to give a heads up that examples might be broken with this PR. I'll file a ticket for that.

jlaine:
@vasilvv @DavidSchinazi I am not sure how to interpret the diverging requirements around `:scheme` and `:path` pseudo-headers between the HTTP/3 draft on one hand and the CONNECT-UDP + WebTransport specs, would you have any insights?

DavidSchinazi:
Yes WebTransport leverages "Extended CONNECT" which was defined for HTTP/2 in [https://datatracker.ietf.org/doc/html/rfc8441](RFC 8441). Since WebTransport requires a SETTING, it's allowed to override any requirements from HTTP/3 if the two specs disagree.

jlaine:
Thanks @DavidSchinazi I hadn't understood that SETTINGS were allowed to introduce behaviour contradicting HTTP/3 requirements!

@silverbucket I've tidied up the PR (rebased + amended the commit message) and will merge it once CI passes.

DavidSchinazi:
Conceptually, SETTINGS allow you to negotiate extensions to HTTP, and those extensions are allowed to change things in HTTP.

jlaine:
> Conceptually, SETTINGS allow you to negotiate extensions to HTTP, and those extensions are allowed to change things in HTTP.

Sure, I understand that SETTINGS enable additional behaviour, I just hadn't realised this additional behaviour could contradict the base specs. Thanks for the clarification!

silverbucket:
Thanks @jlaine and @DavidSchinazi!

PR #216, opened 2021-07-24
Add Chromium/Chrome requirements, example usage

guest271314:


jlaine:
I like the idea of helping users get up and running, so thanks for this! However I think the text needs a bit more work:

- The `--ignore-certificate-errors-spki-list` flag is only needed when you use an *invalid* certificate, such as the one provided in the "tests" directory. If you use a valid certificate (for example the one I use on quic.aiortc.org) you don't need this flag.
- The `--origin-to-force-quic-on` is only needed when you *only* provide an HTTP/3 server. If you are also running an HTTP/1 (or HTTP/2) server with the Alt-Svc header set to advertise HTTP/3 availability, Chrome will switch to HTTP/3 for future requests. This is the case for `quic.aiortc.org`.
- The `--enable-experimental-web-platform-features` flag is only needed for WebTransport, because the specs and Chromium implementation are not yet finalised. For HTTP/3 itself you don't need this flag.

-

guest271314:
Should the list items you posted are succint and informative. Should the text you posted be directly copied to the "Chrome and Chromium usage" section in the PR?

jlaine:
I've reworked the text to integrate the information in the discussion. Thanks a lot the README is now a lot more helpful!

PR #229, opened 2021-09-24
Support the latest draft of HTTP Datagrams

### LINKED ISSUE #228: Support draft-ietf-masque-h3-datagram-04
Please see https://ietf-wg-masque.github.io/draft-ietf-masque-h3-datagram/draft-ietf-masque-h3-datagram.html for the latest draft.

yutakahirano:
This PR contains two changes:
 1. Accept connections with Settings: H3_DATAGRAM=2. Also send
    Settings: H3_DATAGRAM=2. The received and sent settings
    parameters are exposed.
 2. Define H3Capsule which represents the Capsule concept
    defined in [A].

A: https://datatracker.ietf.org/doc/html/draft-schinazi-masque-protocol

Fixes #228.

### DIFF
--- a/src/aioquic/h3/connection.py
+++ b/src/aioquic/h3/connection.py
@@ -1,7 +1,7 @@
 import logging
 import re
 from enum import Enum, IntEnum
-from typing import Dict, FrozenSet, List, Optional, Set
+from typing import Dict, FrozenSet, Iterable, List, Optional, Set
 
 import pylsqpack
 
@@ -80,6 +80,18 @@ class Setting(IntEnum):
     #  Dummy setting to check it is correctly ignored by the peer.
     # https://tools.ietf.org/html/draft-ietf-quic-http-34#section-7.2.4.1
     DUMMY = 0x21
+
+
+class CapsuleType(IntEnum):
+    # Defined in
+    # https://www.ietf.org/archive/id/draft-ietf-masque-h3-datagram-03.html.
+    DATAGRAM = 0xff37a0
+    REGISTER_DATAGRAM_CONTEXT = 0xff37a1
+    REGISTER_DATAGRAM_NO_CONTEXT = 0xff37a2
+    CLOSE_DATAGRAM_CONTEXT = 0xff37a3
+    # Defined in
+    # https://www.ietf.org/archive/id/draft-ietf-webtrans-http3-01.html.
+    CLOSE_WEBTRANSPORT_SESSION = 0x2843
 
 
 class StreamType(IntEnum):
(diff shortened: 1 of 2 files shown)


yutakahirano:
Hi @jlaine, would you take a look at this change?

This is an attempt to support [the latest draft of HTTP Datagrams](https://ietf-wg-masque.github.io/draft-ietf-masque-h3-datagram/draft-ietf-masque-h3-datagram.html). 1. is required to support it in our WebTransport over HTTP/3 server, but 2. is not required (we can parse and encode the capsules at the WebTransport over HTTP/3 server layer). If you think that's better, I'll be happy to modify this PR.

On the other hand, we may want to add more logic in aioquic - For example we can create capsules in H3Connection.send_datagrams. I'm not sure which is better, and would appreciate your feedback.

@DavidSchinazi @vasilvv Can you take a look at this change too? I may misunderstand the draft spec and would appreciate your feedback too.

yutakahirano:
cc: @bashi @nidhijaju

src/aioquic/h3/connection.py
```diff
@@ -1042,23 +1099,23 @@ def _receive_stream_data_uni(
 
     def _validate_settings(self, settings: Dict[int, int]) -> None:
         if Setting.H3_DATAGRAM in settings:
-            if settings[Setting.H3_DATAGRAM] not in (0, 1):
-                raise SettingsError("H3_DATAGRAM setting must be 0 or 1")
+            if settings[Setting.H3_DATAGRAM] not in {0, 1, 2}:
```
DavidSchinazi, line 1102: The change between draft-00 and draft-04 was the setting identifier, not the value. The value is still 0 or 1

src/aioquic/h3/connection.py
```diff
@@ -304,6 +316,35 @@ def validate_trailers(headers: Headers) -> None:
...
+    def decode(data: bytes) -> any:
+        """
+        Returns an H3Capsule representing the given bytes.
+        """
+        buffer = Buffer(data=data)
+        type = buffer.pull_uint_var()
+        length = buffer.pull_uint_var()
+        return H3Capsule(type, buffer.pull_bytes(length))
+
+    def encode(self) -> bytes:
```
DavidSchinazi, line 338: I might be missing something, but where is the code that read/writes the encoded capsules to/from the data stream?

DavidSchinazi (commented):
Thanks for doing this!

src/aioquic/h3/connection.py
yutakahirano: I was thinking about using this in our WebTransport over HTTP/3 server (https://github.com/web-platform-tests/wpt/blob/master/tools/webtransport/h3/webtransport_h3_server.py).

I think it'd be beneficial to have this here because many WebTransport over HTTP/3 servers will use it.

src/aioquic/h3/connection.py
yutakahirano: Thank you, I misunderstood it. I'll revert the change(s).

src/aioquic/h3/connection.py
DavidSchinazi: Oh that makes sense, thanks for clarifying!

yutakahirano:
After talking with David I think I don't need this change. We don't need the full functionality of HTTP/3 datagrams and implementing what's needed on our server directly is probably better. I'm closing this.

PR #230, opened 2021-09-27
Add accessor to sent and received SETTINGS

yutakahirano:
To allow the customer to change behavior based on the sent and
received SETTINGS frames, this change adds accessors to
H3Connection.


yutakahirano:
Hi @jlaine , sorry for bothering you again!

We want to see the sent and received SETTINGS values in a higher layer component, and  hence I'm adding accessors to the values. Waht do you think about this change?

Thanks in advance.

PR #243, opened 2021-12-06
Updated README.rst for latest openssl@3 post install instructions.

michael-observerly:
Updated README.rst to reference latest brew install of openssl@3 post install environment variables (CPPFLAGS and LDFLAGS) link instructions as of 12.2021.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -93,8 +93,8 @@ You will need to set some environment variables to link against OpenSSL:
 
 .. code-block:: console
 
-   $ export CFLAGS=-I/usr/local/opt/openssl/include
-   $ export LDFLAGS=-L/usr/local/opt/openssl/lib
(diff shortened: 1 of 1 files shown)


jlaine:
I understand the @3 but how come the prefix changed too?

michael-observerly:
@jlaine CPPFLAGS are very similar to CFLAGS - however, if you want the flag to be used for _**all languages**_ put it in CPPFLAGS, if it's for a specific language put it in CFLAGS, CXXFLAGS etc.

jlaine:
I understand what `CPPFLAGS` and `LDFLAGS` mean just fine, thanks :)

I'm asking why the `/usr/local` prefix changed to `/opt/homebrew`. What is the standard out-of-the-box prefix for a fresh homebrew installl?

jlaine:
I'm going to close this: the default prefix is still `/usr/local` according to https://brew.sh/

PR #244, opened 2021-12-18
[webtransport] update H3_DATAGRAM value

saitolume:
Update H3_DATAGRAM according to the latest specs.

ref: https://datatracker.ietf.org/doc/html/draft-ietf-masque-h3-datagram#:~:text=H3_DATAGRAM%20%20%7C%200xffd277%20%7C%20This%20Document%20%7C%200

### DIFF
--- a/src/aioquic/h3/connection.py
+++ b/src/aioquic/h3/connection.py
@@ -74,7 +74,7 @@ class Setting(IntEnum):
     MAX_HEADER_LIST_SIZE = 0x6
     QPACK_BLOCKED_STREAMS = 0x7
     NUM_PLACEHOLDERS = 0x9
-    H3_DATAGRAM = 0x276
+    H3_DATAGRAM = 0xffd277
     ENABLE_WEBTRANSPORT = 0x2B603742
 
     #  Dummy setting to check it is correctly ignored by the peer.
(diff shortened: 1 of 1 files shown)


jlaine:
Is this change of constants the *only* required change?

jlaine:
Also could you please fix the linter error by making the hex value uppercase?

saitolume:
> Is this change of constants the only required change?

Yes, I have confirmed that the [sample](https://github.com/GoogleChrome/samples/tree/gh-pages/webtransport) works correctly.

saitolume:
> Also could you please fix the linter error by making the hex value uppercase?

@jlaine  Fixed it! Please check again.

jlaine:
Hm. I'm not convinced that's the only required change (the datagram framing is not what I recall), but let's go.

yutakahirano:
Just FYI, we have some logic in our server to offset the version difference (see [H3ConnectionWithDatagram04](https://github.com/web-platform-tests/wpt/blob/master/tools/webtransport/h3/webtransport_h3_server.py#L39) and `supports_h3_datagram_04` calls in https://github.com/web-platform-tests/wpt/blob/master/tools/webtransport/h3/webtransport_h3_server.py).

PR #251, opened 2022-01-16
Update WebTransport handshake

rlyshw:
/wt endpoint wasn't working. Needed to modify header negotiation to work again. 

Changes mostly come from here:
https://github.com/GoogleChrome/samples/blob/gh-pages/webtransport/webtransport_server.py

### DIFF
--- a/examples/http3_server.py
+++ b/examples/http3_server.py
@@ -13,7 +13,7 @@ import wsproto.events
 import aioquic
 from aioquic.asyncio import QuicConnectionProtocol, serve
 from aioquic.h0.connection import H0_ALPN, H0Connection
-from aioquic.h3.connection import H3_ALPN, H3Connection
+from aioquic.h3.connection import H3_ALPN, H3Connection, Setting
 from aioquic.h3.events import (
     DatagramReceived,
     DataReceived,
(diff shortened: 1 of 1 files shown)


jlaine:
This doesn't look great, if H3Connection needs to be updated for recent protocol changes let's just change it there, not in the examples. Otherwise anybody wanting to use H3Connection will need to apply these hacks. Also I suspect the client suffers from the same problem.

jlaine:
AFAICT most of the changes in this PR are not required. The only addition required for the WebTransport handshake to succeed is adding the `sec-webtransport-http3-draft` header in the server's response. This is done in 24eca50ed4d420da6fdd58f0a6f627623c41b8fc.

jlaine:
FYI, with the changes in version 0.9.19 (just released) the Google Chrome example can be simplified, I've submitted a PR to that effect in https://github.com/GoogleChrome/samples/pull/764

@yutakahirano

yutakahirano:
Thank you!

jlaine:
You're welcome @yutakahirano, it's in my interest to keep your examples as lean as possible, it reduces the noise on the issue tracker :)

Issue #242, opened 2021-11-28
[Test] Indefinite/infinite stream from server to client


guest271314:
There does not appear to be any means to sustain an indefinite/infinite stream (for example, streaming a live jam session; full-length movie or live event; web radio station) using either datagrams or streams using `aioquic`.

Kindly create tests for indefinite/infinite streams using datagrams and streams (where only one (1) write to server is used to start the stream) from server to client.

jlaine:
I missed your PR?

guest271314:
No. You closed the handshake issue which last time I tested on Chromium, did not work using this repository.

I am requesting that you write code testing the use case indefinite/infinite stream from client to server, since you know how your code works, I do not, yet.

jlaine:
OK, feel free to re-open with a PR, I don't understand what you want.

guest271314:
Write a test that streams data infinitely from server to client using a WebTransport stream. 

I tried to stream live audio. The data does not stream continuously. It appears some internal buffer keeps growing (when observing task manager) and the stream does not reach the client.

guest271314:
TL;DR

I can capture system audio output and send to browser using this code as host for Native Messaging

```
#!/usr/bin/env -S python3 -u
# https://github.com/mdn/webextensions-examples/pull/157

import sys
import json
import struct
import argparse
import os
import subprocess
from shlex import split

try:
    # Python 3.x version
    # Read a message from stdin and decode it.
    def getMessage():
        rawLength = sys.stdin.buffer.read(4)
        if len(rawLength) == 0:
            sys.exit(0)
        messageLength = struct.unpack('@I', rawLength)[0]
        message = sys.stdin.buffer.read(messageLength).decode('utf-8')
        return json.loads(message)

    # Encode a message for transmission,
    # given its content.
    def encodeMessage(messageContent):
        encodedContent = json.dumps(messageContent).encode('utf-8')
        encodedLength = struct.pack('@I', len(encodedContent))
        return {'length': encodedLength, 'content': encodedContent}

    # Send an encoded message to stdout
    def sendMessage(encodedMessage):
        sys.stdout.buffer.write(encodedMessage['length'])
        sys.stdout.buffer.write(encodedMessage['content'])
        sys.stdout.buffer.flush()

    while True:
        receivedMessage = getMessage()
        process = subprocess.Popen(split(receivedMessage), stdout=subprocess.PIPE)
        os.set_blocking(process.stdout.fileno(), False)
[...]

PR #262, opened 2022-02-05
Test continuous live streaming

guest271314:
This is a PR for https://github.com/aiortc/aioquic/issues/242.

I do not know how to fix the problem nor write tests for /tests.

What I do know is that the server only sends 28 datagrams.

Steps to reproduce:

- Follow the instructions at https://github.com/aiortc/aioquic/tree/main/examples#chromium-and-chrome-usage then navigate to https://localhost:4433/. I included the script (script.js) in htdocs.

Expected result:

- I write/send one (1) datagram to the server.
- The server writes datagrams to client until I tell it to stop with `writer.close()`

What actually happens:

- The server sends 28 datagrams to client then stops.
- `WebTransportError: Connection lost.` thrown twice.

Screenshot_2022-02-05_09-21-10

What I am actually trying to do is stream media using aioquic and `WebTransport`. For that I need only send one (1) initial datagram telling the server what to stream, then read the datagrams until I call `writer.close()` or `transport.close()`. Very simple.

### DIFF
--- a/examples/htdocs/script.js
+++ b/examples/htdocs/script.js
@@ -1,0 +1,20 @@
+var transport = new WebTransport('https://localhost:4433/wt');
+transport.closed.then(console.log, console.error);
+await transport.ready;
+var { readable, writable } = transport.datagrams;
+var number_of_datagrams_received = 0;
+readable
+  .pipeTo(
+    new WritableStream({
+      write(datagram) {
+        console.log({ number_of_datagrams_received, datagram });
+        ++number_of_datagrams_received;
+      },
+      close() {
+        console.log('closed');
+      },
+    })
+  )
+  .then(console.log, console.error);
+var writer = writable.getWriter();
+await writer.write(new Uint8Array([65]));
--- a/examples/http3_server.py
+++ b/examples/http3_server.py
@@ -27,6 +27,10 @@ from aioquic.quic.events import DatagramFrameReceived, ProtocolNegotiated, QuicE
 from aioquic.quic.logger import QuicFileLogger
 from aioquic.tls import SessionTicket
 
+from random import choice
+from string import digits
+import itertools
+
 try:
     import uvloop
 except ImportError:
@@ -303,7 +307,12 @@ class WebTransportHandler:
                 )
             end_stream = True
         elif message["type"] == "webtransport.datagram.send":
-            self.connection.send_datagram(flow_id=self.stream_id, data=message["data"])
+            print('Try to stream datagrams continuously...')
+            for chunk in iter(lambda: ''.join(choice(digits) for i in range(512)).encode(), b''):
+                if chunk is not None:
+                    self.connection.send_datagram(flow_id=self.stream_id, data=chunk) 
+                    self.transmit()
+            # self.connection.send_datagram(flow_id=self.stream_id, data=message["data"])
         elif message["type"] == "webtransport.stream.send":
             self.connection._quic.send_stream_data(
                 stream_id=message["stream"], data=message["data"]
@@ -315,7 +324,7 @@ class WebTransportHandler:
             )
         if end_stream:
             self.closed = True
-        self.transmit()
+        # self.transmit()
 
 
 Handler = Union[HttpRequestHandler, WebSocketHandler, WebTransportHandler]
--- a/examples/templates/index.html
+++ b/examples/templates/index.html
@@ -31,5 +31,6 @@
             </li>
             <li>There is also an <a href="/httpbin/">httpbin instance</a>.</li>
         </ul>
+    <script type="module" src="/script.js"></script>
     </body>
 </html>


jlaine:
I really don't see how this PR provides anything to the project?

Also if you want to customise the demo behaviour, you don't need to modify the server itself, you can modify the demo:

https://github.com/aiortc/aioquic/blob/4134e6f9952822a4172eef1697d4d68ea2487d5b/examples/demo.py#L103

guest271314:
> I really don't see how this PR provides anything to the project?

The goal is clearly stated in OP: Test continuous, live-streaming; where 1) only one (1) datagram is written to server; (2) datagrams are streamed from server to client indefinitely until `writer.close()` or `transport.close()` are called in the browser.

Currently that does not appear to be possible, either due to idle timeout or some other reason I do not know about.

guest271314:
> Also if you want to customise the demo behaviour, you don't need to modify the server itself, you can modify the demo:

Well, I don't know how to do that to achieve the requirement. You said open a PR in https://github.com/aiortc/aioquic/issues/242. That is what I did here.

I am simply trying to determine if it is possible to live-stream using datagrams or streams. So far I have not located any code which demonstrates that capability using aioquic. I don't know how to explain the use case and where I am stuck any better. I tried explaining here https://groups.google.com/a/chromium.org/g/web-transport-dev/c/UkdFt4aI8RY?pli=1 and here as well https://groups.google.com/a/chromium.org/g/web-transport-dev/c/wpSrznsMJJM/m/fQ3fizBqAgAJ.

guest271314:
> I really don't see how this PR provides anything to the project?
> 
> Also if you want to customise the demo behaviour, you don't need to modify the server itself, you can modify the demo:
> 
> https://github.com/aiortc/aioquic/blob/4134e6f9952822a4172eef1697d4d68ea2487d5b/examples/demo.py#L103

I just tried using code in demo.py - after I changed http3_server.py back to original example code

```
while True:
        message = await receive()
        if message["type"] == "webtransport.datagram.receive":
            print(message)
            for chunk in iter(lambda: ''.join(choice(digits) for i in range(512)).encode(), b''):
                if chunk is not None:
                    await send(
                        {
                            "data": chunk,
                            "type": "webtransport.datagram.send",
                        }
                    )
```

Same result.

Screenshot_2022-02-05_17-48-33

Now, I know I can stream with Python bacause this code works to stream raw PCM using Native Messaging.

```
while True:
        receivedMessage = getMessage()
        process = subprocess.Popen(split(receivedMessage), stdout=subprocess.PIPE)
        os.set_blocking(process.stdout.fileno(), False)
        for chunk in iter(lambda: process.stdout.read(1024 * 1024), b''):
            if chunk is not None:
                encoded = str([int('%02X' % i, 16) for i in chunk])
                sendMessage(encodeMessage(encoded))    
```

Something is limiting the WebTransport stream/datagram output to client. I am more than likely doing something incorrectly. You are the expert. Kindly illuminate.

guest271314:
Importantly, if you are not paying attention to Task Manager - after aioquic stops tranmitting data to the browser - you won't notice that http3_server.py keeps increasing resource consumption by the second. Something ain't right.

Screenshot_2022-02-05_18-41-13
Screenshot_2022-02-05_18-41-33

guest271314:
I created a video of memory consumption of aioquic running the code in this PR (the suggested modification to demo.py). Is a buffer being filled but not transmitted to clent somewhere in your code?

Notice how RSS grows exponentially when the server is run - and continues growing even after WebTransport connection is lost. 

If I am not mistaken this is what is referred to as a memory leak?

https://www.quora.com/What-does-rss-in-the-task-manager-program-of-xubuntu-stand-for

> RSS (Resident Set Size [of Memory]) is that section of the Virtual Memory that's actually in RAM. This number can contain SHR. Hence, the sum of RSS values of two processes need not be the actual RAM they're consuming. So, for example, consider a process that has forked off a child. Until either process changes something, they share the same Resident Set (they just don't know it due to Virtual Memory).

> RSS stands for Resident Set Size. It basically shows the amount of RAM your application is using! If Chrome is using 59 MiB, it means Chrome has taken 59 MiB of your physical memory at this moment.

[aioquic_memory_leak.webm.zip](https://github.com/aiortc/aioquic/files/8010517/aioquic_memory_leak.webm.zip)

guest271314:
I am able to send 16383 datagrams using 

```
 def h3_event_received(self, event: H3Event) -> None:
        if isinstance(event, DatagramReceived):
            for i in range(16384):
               self._http.send_datagram(self._session_id, str([0 for x in range(99)]).encode())
```

What I notice is that RSS increases and remains at the memory consumption size reached when the datagram writing stops. That indicates the data is stored in a buffer somewhere internally, and is never flushed. What is the procedure to actually stream without writing to an intermediate buffer, or flush the buffer after each send?

guest271314:
I'm just telling you your code doesn't work. 

I documented what I tried to make your code work.

Concentrate on your broken code.

jlaine:
> Concentrate on your broken code.

Concentrate on your broken social skills, there might be a reason you got banned by the W3C.

guest271314:
> Concentrate on your broken social skills, there might be a reason you got banned by the W3C.

No. I don't give a damn about W3C or what you consider "social skills". You wouldn't make it 2 seconds where I am from, nor in any of the professionals wher I purvey my wares. Because they, like you, think you are the smartest people in the world while in the meantime your code is broken, with runaway memory from testing stream for a few seconds.

I am not here to make friends. I don't care about the flag you fly or who you think you are.

What is unambiguous is your code has a memory leak. I am just testing aioquic and `WebTransport`.I already achived my requirement multiple ways. `fetch()` with a few lines of PHP is far more reliable than all of the code in your repository to live stream.

Good enough ain't good enough.

Test and fix your broken code.

jlaine:
Let's recap:
- you open issues in a "stream of consciousness" fashion demanding urgent support for your custom projects
- you modify an example and introduce a faulty behaviour (introducing an infinite tight loop isn't going to work)
- you open a pull request which fixes nothing and hurl abuse at the maintainer

I make no claim of being one of "smartest people in the world". On the other hand I do have a demonstrable claim that I've invested thousands of hours into writing a QUIC + HTTP/3 stack, which is freely available to you.

It might be time to revisit your toxic attitude towards opensource projects, this does not end well.

PR #269, opened 2022-03-10
Enable IPv4/IPv6 dual stack support on Windows

msoxzw:
Actually, Windows supports dual stack sockets, but does not enable it by default.

Furthermore, I am curious about the incentive for the use of dual stack, which is inconsistent with [cpython](https://github.com/python/cpython/blob/8714b6fa27271035dd6dd3514e283f92d669321d/Lib/asyncio/base_events.py#L1494-L1502).

### DIFF
--- a/src/aioquic/asyncio/client.py
+++ b/src/aioquic/asyncio/client.py
@@ -1,7 +1,6 @@
 import asyncio
 import ipaddress
 import socket
-import sys
 from contextlib import asynccontextmanager
 from typing import AsyncGenerator, Callable, Optional, cast
 
@@ -11,6 +10,10 @@ from ..tls import SessionTicketHandler
 from .protocol import QuicConnectionProtocol, QuicStreamHandler
 
 __all__ = ["connect"]
+
+# keep compatibility for Python 3.7 on Windows
+if not hasattr(socket, "IPPROTO_IPV6"):
+    socket.IPPROTO_IPV6 = 41
 
 
 @asynccontextmanager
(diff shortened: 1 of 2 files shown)


jlaine:
I'm not 100% sure anymore but I think the rationale for using dual-stack sockets was that we don't know ahead of time what IP protocols we will be dealing with. We may get connection migrations during the lifetime of the connection, or be told to switch to a preferred address which uses a different protocol.

jlaine:
There is a linter error (don't use bare Except) and the test suite fails on Windows / Python 3.7

src/aioquic/asyncio/client.py
```diff
@@ -71,14 +75,15 @@ async def connect(
...
     )
 
-    # connect
+    # explicitly enable IPv4/IPv6 dual stack
     sock = socket.socket(socket.AF_INET6, socket.SOCK_DGRAM)
     try:
         sock.setsockopt(socket.IPPROTO_IPV6, socket.IPV6_V6ONLY, 0)
         sock.bind((local_host, local_port))
-    except:
+    except BaseException:
```
jlaine, line 83: `Exception` not `BaseException`

msoxzw: Socket should be closed when any exceptions occur, so `Exception`, in my view, is not an ideal workaround.

jlaine:
This PR looks great @msoxzw ! I've squashes all the commits and will merge once CI completes.

PR #275, opened 2022-04-05
[tls] better support ALPN byte wire values

msoxzw:
https://bugs.openjdk.java.net/browse/JDK-8264177

### DIFF
--- a/src/aioquic/tls.py
+++ b/src/aioquic/tls.py
@@ -438,11 +438,11 @@ def push_key_share(buf: Buffer, value: KeyShareEntry) -> None:
 
 
 def pull_alpn_protocol(buf: Buffer) -> str:
-    return pull_opaque(buf, 1).decode("ascii")
+    return pull_opaque(buf, 1).decode("latin-1")
 
 
 def push_alpn_protocol(buf: Buffer, protocol: str) -> None:
-    push_opaque(buf, 1, protocol.encode("ascii"))
+    push_opaque(buf, 1, protocol.encode("latin-1"))
 
 
 # PRE SHARED KEY


jlaine:
Ehm no this does not look right. If anything it sounds as though the ALPN values should be handled as bytes.. but this would break backwards compatibility..

msoxzw:
I think that it will break minimal compatibility if has. First, this PR is not intended to represent ALPN values as bytes, but change character encoding about how to convert string to bytes and vice versa. Second, "latin-1", more precisely "ISO_8859-1:1987", ranges from  U+0000 to U+00FF, and has identical code points to "ASCII" when less than U+0080, Unicode when less than U+0100.

jlaine:
I am really not comfortable with forcing latin1 decoding to a string, this is really a hack. If ALPN can contain arbitrary bytes then so be it we will have to make a breaking change.

The breakage should be minimal for most usecases although since for HTTP users should be using the H3_ALPN constant exposed by aioquic. As a convenience and to facilitate the migration we could check the values passed to QuicConfiguration and encode strings to bytes (using ASCII, as currently) and emit a deprecation warning. Bytes would be passed through unchanged.

msoxzw:
Thank you. After I thoroughly read [RFC 7301](https://datatracker.ietf.org/doc/html/rfc7301), I realize my misunderstanding of [ALPN Identification Sequence](https://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml#alpn-protocol-ids), which consist of only octets. Currently, I have not found any RFC and IETF that specify ALPN character encoding.

Therefore, "latin-1" encoding is indeed an ugly hack, which is particularly incompatible with UTF-8, and your proposed solution is much more agreeable and maintainable.

jlaine:
OK thanks for the research, would you mind opening an issue for this so we don't lose track of it?

msoxzw:
Given that CPython has similar issues in [ssl.py](https://github.com/python/cpython/blob/1c8b3b5d66a629258f1db16939b996264a8b9c37/Lib/ssl.py#L565-L575) and [_ssl.c](https://github.com/python/cpython/blob/1c8b3b5d66a629258f1db16939b996264a8b9c37/Modules/_ssl.c#L2064-L2081), this bug could be also created for CPython, and hence, is is advisable to open issue in CPython, aioquic, or both?

PR #280, opened 2022-04-21
[asyncio] optimize QUIC speed on Windows

msoxzw:
The speed of QUIC on Windows is more than 10 times faster when transferring a large file, e.g. 64 MiB.

Test file:
https://gist.github.com/msoxzw/290cca70170ea9f6082c153ce391eb47
https://gist.github.com/msoxzw/567b747e70e2bb69d52586680f0c3a78

Rational:
[`loop.time()`](
https://github.com/python/cpython/blob/f8dc6186d1857a19edd182277a9d78e6d6cc3787/Lib/asyncio/base_events.py#L697-705) calls `time.monotonic()`, but [its time resolution varies from 10 ns to 15 ms depending on operating systems](https://peps.python.org/pep-0418/#monotonic-clocks), the worst of which is Windows. Therefore, using `time.perf_counter()`, a clock with the highest available resolution, basically eliminates time resolution discrepancies among platforms. Additionally, operating systems except Windows is not affected, because [`time.perf_counter()`](https://peps.python.org/pep-0418/#time-perf-counter)  calls  `time.monotonic()` on them.

### DIFF
--- a/src/aioquic/asyncio/protocol.py
+++ b/src/aioquic/asyncio/protocol.py
@@ -1,5 +1,4 @@
 import asyncio
-import time
 from typing import Any, Callable, Dict, Optional, Text, Tuple, Union, cast
 
 from ..quic import events
@@ -58,7 +57,7 @@ class QuicConnectionProtocol(asyncio.DatagramProtocol):
 
         This method can only be called for clients and a single time.
         """
-        self._quic.connect(addr, now=time.perf_counter())
+        self._quic.connect(addr, now=self._loop.time())
         self.transmit()
 
     async def create_stream(
@@ -100,7 +99,7 @@ class QuicConnectionProtocol(asyncio.DatagramProtocol):
         self._transmit_task = None
 
         # send datagrams
-        for data, addr in self._quic.datagrams_to_send(now=time.perf_counter()):
+        for data, addr in self._quic.datagrams_to_send(now=self._loop.time()):
             self._transport.sendto(data, addr)
 
         # re-arm timer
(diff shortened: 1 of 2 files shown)


jlaine:
Wow this looks like an interesting catch, I have to admit I have not had a chance to profile Windows.

One thing worries me though, is that we are calling `self._loop.call_at` on these times, what guarantee do we have that time.perf_counter() is consistent with the event loop's notion of time? What happens if you swap out the event loop (hello uvloop)?

msoxzw:
`time.perf_counter()` becomes system-wide on Windows since Python 3.10, which explains the CI failures.

From my perspective, this issue should be fixed in Python, and there are three ways to resolve it:

1. Wait for https://github.com/python/cpython/issues/88494 to be addressed.
2. Replace "monotonic" with "perf_counter" in [Lib/asyncio/base_events.py](https://github.com/python/cpython/tree/main/Lib/asyncio/base_events.py) on Windows just like many other languages and libraries. As of now, tow lines should be modified: [400th line](https://github.com/python/cpython/blob/254aaa7981d8773658fb14795da5dec888c95f93/Lib/asyncio/base_events.py#L400) and [704th line](https://github.com/python/cpython/blob/254aaa7981d8773658fb14795da5dec888c95f93/Lib/asyncio/base_events.py#L704).
3. Implement `loop.time()` like libuv: simply put,  use CLOCK_MONOTONIC_COARSE if available, otherwise CLOCK_MONOTONIC or equivalent.

jlaine:
Hi @msoxzw I'm kind of confused by this PR. It now only changes `tests/utils.py` so I fail to see what it brings to the table?

msoxzw:
I think it is an upstream issue, so this PR should not be merged or closed until it is fixed in CPython.  I just test whether aioquic functions well or not if "monotonic" is substituted with "perf_counter" when clock resolution is greater than 1 ms, and provide a temporary work-around if someone needs it.

jlaine:
Hi @msoxzw I'm trying to work my way through outstanding PRs and came upon this one.

From re-reading the conversation it sounded as though we were waiting for things to change on the CPython side, do you know if this has happened?

rthalley:
I don't see any sign of the upstream CPython changes happening.

jlaine:
> I don't see any sign of the upstream CPython changes happening.

Should we close this PR?

msoxzw:
I think so. After all, it is an upstream issue.

PR #281, opened 2022-04-27
[asyncio] directly get IPv4-mapped IPv6 address

msoxzw:


jlaine:
Would you mind explaining what this does, I have to admit I have no idea?

msoxzw:
https://man7.org/linux/man-pages/man3/getaddrinfo.3.html
> If hints.ai_flags specifies the AI_V4MAPPED flag, and hints.ai_family was specified as AF_INET6, and no matching IPv6 addresses could be found, then return IPv4-mapped IPv6 addresses in the list pointed to by res.  If both AI_V4MAPPED and AI_ALL are specified in hints.ai_flags, then return both IPv6 and IPv4-mapped IPv6 addresses in the list pointed to by res.  AI_ALL is ignored if AI_V4MAPPED is not also specified.

jlaine:
Could you try rebasing this on top of main please, see #284 ?

msoxzw:
The CI failure only in  "tests / package-wheel (ubuntu-latest, x86_64) (pull_request)" perplexes me, and it is perhaps attributed to operating system itself or CI misconfiguration.

jlaine:
The fact that CI works fine without the patch, and that the failure is *precisely* related to DNS resolution means we definitely have a regression here..

msoxzw:
Nevertheless, I could not reproduce CI failure, especially in this situation:
✔️ tests / test (ubuntu-latest, 3.10) (pull_request)
✔️ tests / test (ubuntu-latest, 3.9) (pull_request)
✔️ tests / test (ubuntu-latest, 3.8) (pull_request)
✔️ tests / test (ubuntu-latest, 3.7) (pull_request)
✔️ tests / package-wheel (ubuntu-latest, i686) (pull_request)
❌️ tests / package-wheel (ubuntu-latest, x86_64) (pull_request)

jlaine:
@msoxzw I'm going to close this, CI consistently fails

msoxzw:
I second your decision. After all, some platforms don't even support IPv4-mapped IPv6 address altogether.

PR #285, opened 2022-04-28
http3_client.py: fixup 2nd and later urls

### LINKED ISSUE #283: http3_client.py sends empty :scheme and :authority in 2nd url
```
python3 examples/http3_client.py https://www.google.com/ 1 2 3 4
```
when it sends 2nd or later request in the connection to www.google.com, `:scheme` and `:authority` are empty.
Some providers like google and facebook allow this, but Cloudflare rejects the connection because it violates 
https://quicwg.org/base-drafts/draft-ietf-quic-http.html#section-4.1.1.1 (and I believe it's correct)

For example this command will hang for a while:
```
python3 examples/http3_client.py https://www.cloudflare.com/ 1 2 3 4
```

junhochoi:
- Check if all urls are in the same host and scheme.
- If url argument is partial (e.g. without https://),
  fill in fields from 1st url.

Fixes #283


jlaine:
Wait up I have PR #284 to fix some new linter errors that sprung up.

jlaine:
OK let's go with your current patch, we can always improve further later.

PR #310, opened 2022-07-15
updated help for "--retry" argument in examples/http3_server.py

Karthikdasari0423:
just changed help comment from "send a retry for new connections" to "the ASGI application as : to send a retry for new connections" for "--retry" argument

### DIFF
--- a/examples/http3_client.py
+++ b/examples/http3_client.py
@@ -366,6 +366,25 @@ async def run(
     else:
         port = 443
 
+    # check validity of 2nd urls and later.
+    for i in range(1, len(urls)):
+        _p = urlparse(urls[i])
+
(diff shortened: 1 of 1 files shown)


jlaine:
I don't understand the new comment, how is --retry related to the ASGI app?

Karthikdasari0423:
i mean to say "--retry" will accept the ASGI application right instead of any port number or ip address

jlaine:
The `--retry` takes no arguments at all, it's just a flag which makes the server sent a RETRY to any new connection.

Karthikdasari0423:
oh okay

Thank you for confirming and i tried with --retry demo:app and it worked,so i thought to  update "--retry" argument help.

PR #311, opened 2022-07-15
docs(readme): HTTP/3 is standardized now

max-wittig:
https://datatracker.ietf.org/doc/html/rfc9114

Since June 6th 2022


jlaine:
The two parts of the sentence seem a bit weird, what do you think of:

`QUIC was standardised in RFC 9000 and HTTP/3  in RFC 9114.`

ahopkins:
Would be nice to hyperlink the RFCs too.

jlaine:
I've updated the PR to include links to the appropriate RFCs. I've also updated the IETF links in the code comments.

Let's see what CI has to say.

max-wittig:
@jlaine Thanks for the changes!

PR #314, opened 2022-07-17
[examples] update httpx demo for recent httpx versions

jlaine:


jlaine:
@tomchristie @forimondmanca Does this look OK?

tomchristie:
Looks right to me.

tomchristie (approved):
All looks correct to me, yup.

examples/httpx_client.py
```diff
@@ -23,22 +22,26 @@
...
-    async def handle_async_request(
-        self,
-        method: bytes,
-        url: Tuple[bytes, bytes, Optional[int], bytes],
-        headers: Headers = None,
-        stream: httpcore.AsyncByteStream = None,
-        extensions: dict = None,
-    ) -> Tuple[int, Headers, httpcore.AsyncByteStream, dict]:
+    async def handle_async_request(self, request: httpx.Request) -> httpx.Response:
+        assert isinstance(request.stream, httpx.AsyncByteStream)
```
tomchristie, line 43: Aside: Interesting code smell that the `httpx` API is forcing you into here.

Could be a useful prompt that we could?? expose `request.sync_stream` and `request.async_stream` accessors. These would return the correct type, or raise `RuntimeError` if used in the wrong context.

Don't feel strongly enough about this to do anything about it right now, but... interesting thought.

Pointer to the same style within `https`... https://github.com/encode/httpx/blob/master/httpx/_client.py#L1716-L1724

PR #322, opened 2022-08-19
Fix typos

kianmeng:
Found via these command, `codespell . -L quicly`.


jlaine:
Wow this is really neat. Would you consider a followup PR that runs `codespell` as part of our CI linter run?

PR #332, opened 2022-09-17
[logger] log packet_dropped with reason header_parse_error

pyfisch:
Some servers add all zero padding to UDP datagrams during the handshake.

See also <https://github.com/quicwg/base-drafts/issues/3333>, one server with padding outside the QUIC packet is cloudflare/quiche.


jlaine:
OK this looks good. Let me get CI back into shape and I'll loop back to this.

PR #333, opened 2022-09-27
Update httpx example to support latest httpcore/httpx versions

madkinsz:
Updates the httpx client transport example so the basic case works again. 

- The `httpcore` imports are in `httpx` now. 
- The request handler uses actual `Request` and `Response` types instead of tuples.

Tested with the examples from the README

```
❯ python examples/http3_server.py --certificate tests/ssl_cert.pem --private-key tests/ssl_key.pem
2022-09-26 21:52:40,577 INFO quic [acf386387b368251] ALPN negotiated protocol h3
2022-09-26 21:52:40,578 INFO quic [acf386387b368251] HTTP request GET /
2022-09-26 21:52:40,578 INFO quic [acf386387b368251] HTTP request GET /style.css
2022-09-26 21:52:40,585 INFO quic [acf386387b368251] Connection close received (code 0x0, reason )
```

```
❯ python examples/httpx_client.py --ca-certs tests/pycacert.pem -i https://localhost:4433/
2022-09-26 21:53:52,732 INFO quic [7f3cd4a02a7f8d16] ALPN negotiated protocol h3
2022-09-26 21:53:52,732 INFO client New session ticket received
2022-09-26 21:53:52,732 INFO quic [7f3cd4a02a7f8d16] Duplicate CRYPTO data received for epoch Epoch.HANDSHAKE
2022-09-26 21:53:52,733 INFO quic [7f3cd4a02a7f8d16] Duplicate CRYPTO data received for epoch Epoch.ONE_RTT
2022-09-26 21:53:52,736 INFO client Received 1276 bytes in 0.0 s (7.280 Mbps)
2022-09-26 21:53:52,736 INFO quic [7f3cd4a02a7f8d16] Connection close sent (code 0x0, reason )
```

### DIFF
--- a/examples/httpx_client.py
+++ b/examples/httpx_client.py
@@ -28,16 +28,16 @@ logger = logging.getLogger("client")
 class H3Transport(QuicConnectionProtocol, httpx.AsyncHTTPTransport):
     def __init__(self, *args, **kwargs):
         QuicConnectionProtocol.__init__(self, *args, **kwargs)
+
+        # TODO: We must initialize the HTTPTransport but it is not clear
+        #       which arguments belong to it and which are for the other base
         httpx.AsyncHTTPTransport.__init__(self)
 
         self._http = H3Connection(self._quic)
         self._read_queue: Dict[int, Deque[H3Event]] = {}
         self._read_ready: Dict[int, asyncio.Event] = {}
 
-    async def handle_async_request(
-        self,
-        request: Request
-    ) -> Response:
+    async def handle_async_request(self, request: Request) -> Response:
         stream_id = self._quic.get_next_available_stream_id()
         self._read_queue[stream_id] = deque()
         self._read_ready[stream_id] = asyncio.Event()
@@ -47,7 +47,7 @@ class H3Transport(QuicConnectionProtocol, httpx.AsyncHTTPTransport):
             stream_id=stream_id,
             headers=[
                 (b":method", request.method.encode()),
-                (b":scheme", request.url.raw_scheme), 
+                (b":scheme", request.url.raw_scheme),
                 (b":authority", request.url.raw_host),
                 (b":path", request.url.raw_path),
             ]
@@ -70,9 +70,13 @@ class H3Transport(QuicConnectionProtocol, httpx.AsyncHTTPTransport):
             self._receive_response_data(stream_id, stream_ended)
         )
 
-        return Response(status_code=status_code, headers=headers, stream=response_stream, extensions={
+        return Response(
+            status_code=status_code,
+            headers=headers,
+            stream=response_stream,
+            extensions={
                 "http_version": b"HTTP/3",
-            }
+            },
         )
 
     def http_event_received(self, event: H3Event):
--- a/examples/httpx_client.py
+++ b/examples/httpx_client.py
@@ -48,7 +48,7 @@ class H3Transport(QuicConnectionProtocol, httpx.AsyncHTTPTransport):
             headers=[
                 (b":method", request.method.encode()),
                 (b":scheme", request.url.raw_scheme),
-                (b":authority", request.url.raw_host),
+                (b":authority", request.url.netloc),
                 (b":path", request.url.raw_path),
             ]
             + [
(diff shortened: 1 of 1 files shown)


madkinsz:
😮‍💨 that's 20 minutes down the drain. This looks like a slightly different implementation of the already open changes in https://github.com/aiortc/aioquic/pull/314/ — feel free to close if you wish.

It'd be great to get this updated though :)

examples/httpx_client.py
```diff
@@ -47,18 +46,18 @@ async def handle_async_request(
         self._http.send_headers(
             stream_id=stream_id,
             headers=[
-                (b":method", method),
-                (b":scheme", url[0]),
-                (b":authority", url[1]),
-                (b":path", url[3]),
+                (b":method", request.method.encode()),
+                (b":scheme", request.url.raw_scheme), 
+                (b":authority", request.url.raw_host),
```
madkinsz, line 51: Should this be `netloc` or `host`?

tomchristie: I'd need to drag through to specs to give you a proper pointer, but... it should be `netloc`.

Same as the `Host` header in HTTP/1.1... https://github.com/encode/httpx/blob/8152c4facd0f71e0f376287e41a0810a60fec9c6/httpx/_client.py#L547

tomchristie:
> This looks like a slightly different implementation of the already open changes in https://github.com/aiortc/aioquic/pull/314 — feel free to close if you wish. I think this addresses some additional required changes.

Do you want to review the other PR? Which should we be accepting?

madkinsz:
I believe the other will fail due to `AsyncBaseTransport._pool` being undefined. See https://github.com/aiortc/aioquic/pull/333/files#diff-99cc82566217a40c4209ebcc437bd71aa43c67c46aecce11d0fd8a8605236526R32-R34

The primary other difference is that they define their own version of `AsyncResponseStream`. I think it'll probably be easier to port into httpx if we use httpx's type 🤷.

examples/httpx_client.py
```diff
@@ -23,22 +25,19 @@
 logger = logging.getLogger("client")
 
 
-class H3Transport(QuicConnectionProtocol, httpcore.AsyncHTTPTransport):
+class H3Transport(QuicConnectionProtocol, httpx.AsyncHTTPTransport):
```
tomchristie, line 28: The base class for custom Async transports is `AsyncBaseTransport`. You'll want to subclass that, rather than `AsyncHTTPTransport`.

You don't need to call `__init__` for it, as it's (effectively) just an interface.

madkinsz: Ah! That's helpful. Then I bet the other example is working fine.

examples/httpx_client.py
```diff
@@ -9,8 +9,10 @@
 from typing import AsyncIterator, Deque, Dict, Optional, Tuple, cast
 from urllib.parse import urlparse
 
-import httpcore
+import httpx
 from httpx import AsyncClient
+from httpx._transports.default import AsyncResponseStream
+from httpx._models import Request, Response
```
jlaine, line 15: This looks as though we're depending on private API, which is likely to break at any time or am I mistaken?

tomchristie: Yes,well caught @jlaine.
You should import from the `httpx` rather than the (private) module space.

You want `httpx.Request` and `httpx.Response`.

The `AsyncResponseStream` class is private. Instead you'll want your own response stream that subclasses `httpx.AsyncByteStream()`...

```python
class AsyncResponseStream(AsyncByteStream):
    def __init__(self, response_stream: typing.AsyncIterable[bytes]):
        self._response_stream = response_stream

    async def __aiter__(self) -> typing.AsyncIterator[bytes]:
        # you might want to wrap any exceptions you can throw and coerce them into `httpx` exceptions.
        async for part in self._response_stream:
            yield part

    async def aclose(self) -> None:
        ... # hook into any close behaviour you have
```

jlaine:
OK I've merged #314 so I'm going to close this.

madkinsz:
Thanks @jlaine !

PR #338, opened 2022-11-02
Get next available stream ID more efficiently

msoxzw:
The [`itertools.count()`](https://docs.python.org/3/library/itertools.html#itertools.count) not only satisfies [stream ID requirement](https://www.rfc-editor.org/rfc/rfc9000.html#name-stream-types-and-identifier), but also runs faster, because it remembers current stream ID and is implemented by C.


jlaine:
I don't think this is correct, you are changing the state of the connection on every call, which is not the current behaviour?

jlaine:
Closing in favour of #403

PR #340, opened 2022-11-03
Implement address validation token for future connections for client

msoxzw:
https://www.rfc-editor.org/rfc/rfc9000.html#name-address-validation-for-futu

### DIFF
--- a/src/aioquic/quic/connection.py
+++ b/src/aioquic/quic/connection.py
@@ -5,7 +5,18 @@ from collections import deque
 from dataclasses import dataclass
 from enum import Enum
 from functools import partial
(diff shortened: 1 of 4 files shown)


src/aioquic/quic/connection.py
```diff
@@ -240,6 +251,7 @@ def __init__(
         retry_source_connection_id: Optional[bytes] = None,
         session_ticket_fetcher: Optional[tls.SessionTicketFetcher] = None,
         session_ticket_handler: Optional[tls.SessionTicketHandler] = None,
+        token_handler: Optional[Callable] = None,
```
jlaine, line 254: Can we do better than `Callable` here?

PR #350, opened 2023-01-06
Use function parameters instead of global arg values

eruvanos:
Clean up example, which uses global variables and ignores parameters passed to the main function.


jlaine:
Absolutely, nice catch. I will rebase this once I get CI back up.

PR #352, opened 2023-01-11
[logger] log version_information events

pyfisch:
Use more events from qlog draft for easier debugging.

### DIFF
--- a/src/aioquic/quic/connection.py
+++ b/src/aioquic/quic/connection.py
@@ -802,7 +802,18 @@ class QuicConnection:
                 common = [
                     x for x in self._configuration.supported_versions if x in versions
                 ]
-                if not common:
+                chosen_version = common[0] if common else None
+                if self._quic_logger is not None:
+                    self._quic_logger.log_event(
+                        category="transport",
+                        event="version_information",
+                        data={
+                            "server_versions": versions,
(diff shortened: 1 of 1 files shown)


jlaine:
This looks great, would you have a pointer to the corresponding specs?

pyfisch:
> It looks as though there are several other cases where we should be logging this event?

Not sure, which cases do you think are missing?

src/aioquic/quic/connection.py
```diff
@@ -789,7 +789,18 @@ def receive_datagram(self, data: bytes, addr: NetworkAddress, now: float) -> Non
...
-                if not common:
+                chosen_version = common[0] if common else None
+                if self._quic_logger is not None:
+                    self._quic_logger.log_event(
+                        category="transport",
+                        event="version_information",
+                        data={
+                            "server_versions": versions,
+                            "client_versions": self._configuration.supported_versions,
+                            "chosen_version": chosen_version,
```
jlaine, line 800: I don't think passing `None` here is valid.

jlaine:
> > It looks as though there are several other cases where we should be logging this event?
> 
> Not sure, which cases do you think are missing?

It looks like we should have `server_versions` ?

jlaine:
Let's merge this as-is, we can always iterate.

PR #356, opened 2023-02-07
Update the value of H3_DATAGRAM to RFC9297

KershawChang:
Since RFC9297has been published, we might want to update the value of `H3_DATAGRAM` to `0x33`.


jlaine:
Hi @KershawChang and thanks for this! Is this the only constant which needs to be changed?

PR #373, opened 2023-05-08
build abi3 wheels

kdashg:
- change c extensions to use the limited ABI
 - update `setup.py` to build abi3 wheels
 - add Python 3.11 to build matrix
 - rebased 2023may08 from #342

### DIFF
--- a/src/aioquic/_buffer.c
+++ b/src/aioquic/_buffer.c
@@ -15,6 +15,8 @@ typedef struct {
     uint8_t *pos;
 } BufferObject;
 
+static PyObject *BufferType;
+
 #define CHECK_READ_BOUNDS(self, len) \
     if (len < 0 || self->pos + len > self->end) { \
         PyErr_SetString(BufferReadError, "Read out of bounds"); \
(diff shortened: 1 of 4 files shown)


kdashg:
#342 appears to have stalled, so here's a rebase!

kdashg:
@jlaine PTAL!

jlaine:
I'd love to get this merged, but CI seems to be in a weird place right now..

src/aioquic/_buffer.c
```diff
@@ -5,6 +5,12 @@
 
 #define MODULE_NAME "aioquic._buffer"
 
+// https://foss.heptapod.net/pypy/pypy/-/issues/3770
```
jlaine, line 8: It looks like this issue was fixed, can we drop this?

jlaine:
OK I've made a couple of changes:

- `wheel` needed to be added to build requirements in `pyproject.toml` since we import it from our `setup.py`
- the workaround for pypy was not needed anymore
- the imports in `setup.py` were not sorted correctly

Issue #374, opened 2023-05-17
Non-ASCII ALPN


martinthomson:
https://github.com/aiortc/aioquic/blob/5757081c8d4f4675890a447f211c60b367eb93a1/src/aioquic/tls.py#L437 decodes ALPN values as ASCII strings.  This is usually OK, but not always.  Grease for ALPN, as defined in [RFC 8701](https://datatracker.ietf.org/doc/html/rfc8701), includes ALPN values that are not safe to decode as ASCII.  aioquic as a server doesn't like those and will crash if a client sends ALPN like this.  It probably shouldn't.

One fix is perhaps trivial: catch the decode error when adding to the list of extension values and just skip those.  Real ALPN values are likely to be ASCII strings for now, so the only problem this might create is that people won't be able to configure aioquic with a non-ASCII protocol name.

<details>
<summary>Stack</summary>

```
handle: <Handle _SelectorDatagramTransport._read_ready()>
Traceback (most recent call last):
  File "/usr/lib64/python3.11/asyncio/events.py", line 80, in _run
    self._context.run(self._callback, *self._args)
  File "/usr/lib64/python3.11/asyncio/selector_events.py", line 1169, in _read_ready
    self._protocol.datagram_received(data, addr)
  File "/home/jesup/src/mozilla/inbound_merge/obj-debug/_virtualenvs/wpt/lib/python3.11/site-packages/aioquic/asyncio/server.py", line 149, in datagram_received
    protocol.datagram_received(data, addr)
  File "/home/jesup/src/mozilla/inbound_merge/obj-debug/_virtualenvs/wpt/lib/python3.11/site-packages/aioquic/asyncio/protocol.py", line 135, in datagram_received
    self._quic.receive_datagram(cast(bytes, data), addr, now=self._loop.time())
  File "/home/jesup/src/mozilla/inbound_merge/obj-debug/_virtualenvs/wpt/lib/python3.11/site-packages/aioquic/quic/connection.py", line 980, in receive_datagram
    is_ack_eliciting, is_probing = self._payload_received(
                                   ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jesup/src/mozilla/inbound_merge/obj-debug/_virtualenvs/wpt/lib/python3.11/site-packages/aioquic/quic/connection.py", line 2260, in _payload_received
    frame_handler(context, frame_type, buf)
[...]

PR #375, opened 2023-05-17
Catch errors trying to read greased ALPN values as ASCII

jesup:
This works around the problem reported in issue #374


jlaine:
Hi! Thanks for looking into this.

A couple of points:
- I'm not going to merge any PRs which don't carry a unit test. You're addressing a legitimate problem, it's worth capturing the situation in a test 😃
- The try/catch seems to be at a very generic level, meaning it will silently swallow all kinds of errors. I'd appreciate some insights into why we're doing it here and not for example in more ALPN-specific code?

jlaine:
No response from submitter, closing.

PR #376, opened 2023-06-08
examples: update ports and ALPN for DoQ examples to the current RFC

Payne-X6:
The RFC for DoQ ([RFC 9250](https://datatracker.ietf.org/doc/html/rfc9250)) has been out for some time. I've only updated the ALPN and default port in this PR, the rest with the RFC seems to fit within the scope of this example.

### DIFF
--- a/examples/README.rst
+++ b/examples/README.rst
@@ -101,15 +101,22 @@ DNS over QUIC
 By default the server will use the `Google Public DNS`_ service, you can
 override this with the ``--resolver`` argument.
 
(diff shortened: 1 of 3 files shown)


jlaine:
Thanks for this, it looks good and I tested it against some servers from:

https://cln.io/blog/list-of-quic-dns-servers/

jlaine:
The one downside is that the server example cannot be run as an unprivileged user anymore!

PR #378, opened 2023-06-08
Fix ssl.match_hostname deprecation and various improvements

### LINKED ISSUE #368: Find a replacement for deprecated `ssl.match_hostname`
When connecting to a server, we rely on the standard library's [ssl.matchhostname](https://docs.python.org/3/library/ssl.html#ssl.match_hostname) method to check the certificate presented by the server is valid for the hostname we connected to. Howver this method has been deprecated since Python 3.7 so we need to find a suitable replacement.

So far I have found:

- urllib3 [vendors this method](https://github.com/urllib3/urllib3/blob/main/src/urllib3/util/ssl_match_hostname.py ) into its code.  This does seem like a great option as it puts the burden of maintaining a critical piece of code on us.
- OpenSSL has an [X509_check_host](https://www.openssl.org/docs/man1.1.1/man3/X509_check_host.html) method. AFAICT this is not accessible from pyOpenSSL.

Ousret:
Hello,

First, thank you for the work you have put together in this project, it is appreciated.

I have been working on experimental support for HTTP/3 over QUIC in urllib3 https://github.com/urllib3/urllib3/pull/3030
And I noted some tiny things that could be improved.

Close #368 

**In this PR**:

- (i) Make certifi an optional dependency
- (ii) Backport the proven match_hostname from urllib3 to our convenience
- (iii) Mimic load_default_locations() from SSL stdlib CPython and truststore.openssl

**Explanations:**

- (i) Dependency footprint matters and certifi is more or less not really aioquic responsibility.
- (ii) Fix deprecation.
- (iii) I had some different behavior with default context (TCP) and default context (QUIC), urllib3 default context does call load_default_locations() when nothing was given (cadata, cafile, ...)

**Other:**

Ideally, this project should require only `pylsqpack`, but we both know that isn't going to happen until CPython ships with compatible OpenSSL QUIC support... which OpenSSL does yet support fully as of today.
So yes, we shall remain patient to make this lighter. The alternative would be vendor dependencies but at a high maintenance burden cost.


jlaine:
Thanks a *lot* for this. match_hostname has been a thorn in my side for a while. I'm a bit sad we have to vendor code though, couldn't this be hoisted to pyopenssl, cryptography or some other project?

src/aioquic/__init__.py
```diff
@@ -1 +1 @@
-__version__ = "0.9.20"
+__version__ = "0.9.21"
```
jlaine, line 1: Please don't touch the version

jlaine:
If the code for hostname matching was taken from another project, can you please isolate it in a file to make easier to keep in sync? You will also need to include the test suite for this code.

Rebasing your code on top of main should fix the build errors on Windows.

jlaine:
After further digging I'm now convinced that backporting match_hostname is not the way to go, and instead we should use https://github.com/pyca/service-identity as suggested by @GalaxySnail in #368 .

Rationale:
- match_hostname implements outdated practices, e.g. it uses the `commonName` whereas all browsers have switched to using `subjectAltName` only.
- maintaining a correct implementation of hostname checks ourselves is too much of a burden and any mistake here is a security risk

Ousret:
Thank you for the review. I can see a divergence between what you foresee and what I do.

> If the code for hostname matching was taken from another project, 

It is in the description of the PR. urllib3.

> You will also need to include the test suite for this code.

Not necessarily. We extract it as is and it is proven enough given the distribution range. We should (more) care about the integration test.

> I'm a bit sad we have to vendor code though, couldn't this be hoisted to pyopenssl, cryptography or some other project?

asking a critical project to do this is a challenge in itself.

> After further digging I'm now convinced that backporting match_hostname is not the way to go, and instead we should use https://github.com/pyca/service-identity as suggested by @GalaxySnail in https://github.com/aiortc/aioquic/issues/368 .

It seems like the best move going forward, but I don't think you see the big picture here. `aioquic` should be a leaf dependency and should care about not having much more dependencies.
This matter shouldn't be of `aioquic` but rather of HTTP client (higher) level client.
All of this will go away when the QUIC layer will be implemented in Python directly through the usual wrap socket, the focus will be the h3 protocol and lsqpack logic.

> match_hostname implements outdated practices, e.g. it uses the commonName whereas all browsers have switched to using subjectAltName only.

Yes and no. I did not put it as the default immediately because it would surprise your users. urllib3 does, in fact, rely on subjectAltName only by default but allows to use of the commonName if desired.

> maintaining a correct implementation of hostname checks ourselves is too much of a burden and any mistake here is a security risk

You are not alone. urllib3 is well-backed and the hostname checks are not at risk (so far).

In the current state, it is hard to see a high-level HTTP client use `aioquic` to serve h3 capabilities soon.

Finally, I made numerous links to PR, fork (qh3), project, and research that you may consult (or not).

Regards,

PR #381, opened 2023-06-16
[tls] Match server_name against SAN IPAddress

XeCycle:
This simple change "works for me"; I think it could be useful to some, and worth adding if it does not introduce security holes, although I'm not 100% confident about that.  Do you prefer this as-is, or some sort of opt-in switch for this part?

### DIFF
--- a/src/aioquic/tls.py
+++ b/src/aioquic/tls.py
@@ -226,6 +226,8 @@ def verify_certificate(
                 for name in ext.value:
                     if isinstance(name, x509.DNSName):
                         subjectAltName.append(("DNS", name.value))
+                    elif isinstance(name, x509.IPAddress):
+                        subjectAltName.append(("IP Address", str(name.value)))
 
         try:
             ssl.match_hostname(


jlaine:
Can you please quote the RFC or appropriate spec for this?

We're also going to need an explicit unit test for this

jlaine:
RFC 6125 explicitly considers matching by IP Address to be "out of scope", but I don't know if a subsequent RFC clarifies the expected behaviour:

https://www.rfc-editor.org/rfc/rfc6125.html#section-1.7.2

XeCycle:
[RFC 9001, Section 4.4](https://www.rfc-editor.org/rfc/rfc9001.html#name-peer-authentication) says "for example RFC 2818" which is the old HTTP-over-TLS, although the text does not seem very normative.  The linked [RFC 2818](https://www.rfc-editor.org/rfc/rfc2818.html) simply says "the iPAddress subjectAltName must be present in the certificate and must exactly match the IP in the URI".

There's a discrepancy here: HTTP specifies and uses URI, but QUIC, at the layer below HTTP/3, does not speak URI.  So this text can only be taken as an "example", since it apparently cannot be applied as-is.

So I'm not sure what's the IETF opinion on this.  However this seems to be a common practice: CloudFlare, on 1.1.1.1, both TCP 443 and UDP 443, serves with their certificate containing SAN IP:1.1.1.1.  I did not try any HTTP/3 clients, but as for HTTP 1 or 2 over TCP TLS, these clients accept that cert by default:
- Firefox
- Chromium
- curl (on Linux using OpenSSL)
- PowerShell Invoke-WebRequest
- iOS Safari

Just to be clear, I don't think either CloudFlare or these clients are authoritative on this topic; but they give me a feeling that support for this feature is more frequently implemented than not.

Btw when digging those RFCs I noticed a missing piece: [RFC 5280, Section 4.2.1.10](https://www.rfc-editor.org/rfc/rfc5280#section-4.2.1.10) specifies that "Name Constraints" can be given in CIDR ranges, and (as I understand it) should be checked; I think this check is not implemented in this PR. I'm unsure how useful it is; [this SO question](https://serverfault.com/q/975838) implies Chrome does implement the check.  Oh this is so complex I need some more time if we want this constraint implemented together...

jlaine:
Ok fair enough I think I'm convinced the usecase is legitimate, thanks for the research!

Implementation-wise, I'm currently considering switching subject validation to `service-identity` as `match_hostname` is deprecated, see #368. I see that this package provides support both for validating FQDNs and IP addresses against the SAN.

XeCycle:
> considering switching subject validation to `service-identity`

Yes, it covers matching of the IP.  But it seems name constraints are not covered; I'm feeling that this may be a limitation in the API, because such constraints come from some issuer up the trust chain, but if we take only the end certificate as input we cannot tell if there's a constraint to apply.

> how do you plan to pass an IP address to the TLS stack?

I create `QuicConfiguration(server_name=host)` which is the IP in string format, thus I said `server_name` in my title.

jlaine:
Superseeded by #390

PR #392, opened 2023-07-20
Change DataGram size to allowed Limit

Arvind2222:
* 1280 to 1200 Ref #https://www.rfc-editor.org/rfc/rfc9000.html#name-datagram-size


Arvind2222:
@jlaine Could you please review

jlaine:
Since we've been bitten by the distinction between payload size and IP packet size, a comment seems necessary here.

jlaine:
As expected this breaks test pretty much all over the place, they need fixing..

jlaine:
Let's address this in #400

PR #393, opened 2023-07-21
Add StopSendingReceived QUIC event

bashi:
Section 2.4 of RFC9000 says that an application protocol can request to be informed of state changes on streams. Even though the state machines described in section 3 don't explicitly have state transitions regarding STOP_SENDING, an application protocol sometimes needs to know the reception of STOP_SENDING.

For example, WebTransport Web API defines the [cancel](https://w3c.github.io/webtransport/#webtransportreceivestream-cancel) algorithm that sends a STOP_SENDING. To write a test for the algorithm, a server needs to detect whether the peer sends a STOP_SENDING.

The `StopSendingReceived` event can be used for such purposes.

### DIFF
--- a/docs/quic.rst
+++ b/docs/quic.rst
@@ -44,6 +44,9 @@ Events
     .. autoclass:: PingAcknowledged
         :members:
 
+    .. autoclass:: StopSendingReceived
+        :members:
+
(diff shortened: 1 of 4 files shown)


jlaine:
This is missing unit tests

PR #400, opened 2023-08-29
Max packet size

ihlar:
This PR adds the possibility of configuring the maximum QUIC packet size. Increasing the packet size to make full use of MTU has beneficial performance improvements. It also simplifies future potential implementation of Path MTU Discovery.


src/aioquic/quic/recovery.py
```diff
@@ -15,8 +15,8 @@
 
 # congestion control
 K_MAX_DATAGRAM_SIZE = 1280
```
jlaine, line 17: Do we still need this constant?

src/aioquic/quic/connection.py
```diff
@@ -304,6 +304,8 @@ def __init__(
             frame_type=QuicFrameType.MAX_STREAMS_UNI, name="max_streams_uni", value=128
         )
         self._loss_at: Optional[float] = None
+        assert configuration.max_packet_size >= 1200, "The smallest allowed maximum packet size is 1200 bytes"
```
jlaine, line 307: Can we move this assertion up to the top of the constructor with the other ones?

src/aioquic/quic/configuration.py
```diff
@@ -46,6 +46,11 @@ class QuicConfiguration:
     Connection-wide flow control limit.
     """
 
+    max_packet_size: int = 1280
+    """
+    Maximum size of quic packets.
```
jlaine, line 51: QUIC in capitals, and it feels as though we should be more explicit about what we mean by "packet size". (without UDP overhead, etc..)

src/aioquic/quic/recovery.py
jlaine: Nope

jlaine:
From reading the RFCs it looks as though we should be using `max_datagram_size` to be consistent with their terminology.

src/aioquic/quic/connection.py
```diff
@@ -539,8 +545,11 @@ def datagrams_to_send(self, now: float) -> List[Tuple[bytes, NetworkAddress]]:
             builder.max_flight_bytes = (
                 self._loss.congestion_window - self._loss.bytes_in_flight
             )
-            if self._probe_pending and builder.max_flight_bytes < PACKET_MAX_SIZE:
-                builder.max_flight_bytes = PACKET_MAX_SIZE
+            if (
```
jlaine, line 548: I am not sure this is correct. I think this should be exactly 1200 bytes.

jlaine:
On the whole this PR is now in a good place. I'm still wondering about naming though, as the `max_datagram_size`:

- Only applies to what we *send*.
- Is very close to the `max_udp_datagram_size` QUIC setting, which controls what we are willing to *receive*. It is not currently possible to set this via QuicConfiguration.

=> Should our new setting reflect the fact it applies to *egress*? Like `max_egress_datagram_size`?

Arvind2222:
@jlaine Was doing a retro, I think if we can keep these Configs in good place since the RFC is still new and might change in future. Your call !

PR #402, opened 2023-09-12
Added several congestion control algorithms

Aperence:
Hi,

I've added the following congestion control algorithms to aioquic : cubic, bbrv1, bbrv2.

Cubic was tested, while bbrv1/2 was only implemented, and verified graphically, but no formal verification was done.

I also ran an interop test with cubic to ensure that the implementation wasn't broken after my modifications, here are the [results](https://github.com/aiortc/aioquic/files/12583057/results_cubic.zip). Overall, an average gain of 67kbps was observed for crosstraffic, and the implementation seemed to work still fine (Note: quinn seems to have a problem with cubic, but it is surely due to the way they handle the acks that doesn't work well with cubic).

Also I added hystart++ as a choice for the slow start algorithm.

To choose a congestion control algorithm, just fill the parameter in the config with the congestion control you want to use. For example:
```py
from aioquic.quic.congestion.cubic import CubicCongestionControl
from aioquic.quic.congestion.reno import RenoCongestionControl
from aioquic.quic.congestion.bbr import BBRCongestionControl
from aioquic.quic.congestion.bbr2 import BBR2CongestionControl
from aioquic.quic.congestion.slow_starts import HyStart
from aioquic.quic.configuration import QuicConfiguration
import argparse

args = argparse.ArgumentParser().parse_args()

if (args.cca == "reno"):
    cca = RenoCongestionControl(slow_start=HyStart())
elif (args.cca == "cubic"):
    cca = CubicCongestionControl(reno_friendly_activated=args.reno_friendly)
elif (args.cca == "bbr"):
    cca = BBRCongestionControl()
elif (args.cca == "bbr2"):
    cca = BBR2CongestionControl()
else:
    print("invalid cca")
    exit(1)

conf = QuicConfiguration(alpn_protocols=["quic-testing"], is_client=True, congestion_control=cca)
```

Here are example on how cubic and bbr perform for the growth of cwnd:
- cubic : cubic
- bbr : bbr


jlaine:
I really appreciate the work that has gone into this! However it's a HUGE pull request, and I think we should focus on adding *one* new CC method at a time, with extensive unit tests. From what you've observed what would be the best algorithm to add first?

jlaine:
Great, let's focus on `cubic` then. Maybe a good way to proceed would be to leave this PR as-is so we don't break it and start a fresh PR which only adds `cubic` so we can polish it. There is going to be some adjustment needed because the congestion controller now gets handed a `max_datagram_size` argument instead of relying on the old PACKET_MAX_SIZE.

Aperence:
I’ve opened a new PR (#415) with the needed changes normally

jlaine:
Hi @Aperence, if it's OK with you I suggest we close this PR and if you want to add another algorithm you open a specific PR?

Aperence:
Yes, it seems fine to me :+1:

PR #415, opened 2023-11-07
Support for cubic

Aperence:
See #402  for the details.

This PR provide support for cubic, and support for selecting the maximum datagram size.

Exemple of usage : 
```py
from aioquic.quic.congestion.cubic import CubicCongestionControl
from aioquic.quic.congestion.reno import RenoCongestionControl
from aioquic.quic.congestion.slow_starts import HyStart

if (args.cca == "reno"):
    cca = RenoCongestionControl(max_datagram_size=K_MAX_DATAGRAM_SIZE, callback=callback, slow_start=HyStart(max_datagram_size=K_MAX_DATAGRAM_SIZE))
elif (args.cca == "cubic"):
    cca = CubicCongestionControl(max_datagram_size=K_MAX_DATAGRAM_SIZE, callback=callback, reno_friendly_activated=args.reno_friendly)
else:
    exit(1)

conf = QuicConfiguration(alpn_protocols=["quic-testing"], is_client=True, congestion_control=cca)
```

### DIFF
--- a/src/aioquic/quic/configuration.py
+++ b/src/aioquic/quic/configuration.py
@@ -30,7 +30,7 @@ class QuicConfiguration:
     """
     The name of the congestion control algorithm to use.
 
-    Currently supported algorithms: `"reno"`.
+    Currently supported algorithms: `"reno", `"cubic"`.
     """
 
     connection_id_length: int = 8
--- a/src/aioquic/quic/congestion/base.py
+++ b/src/aioquic/quic/congestion/base.py
@@ -1,5 +1,5 @@
 import abc
-from typing import Dict, Iterable, Optional, Protocol
+from typing import Any, Dict, Iterable, Optional, Protocol
 
 from ..packet_builder import QuicSentPacket
 
@@ -21,7 +21,7 @@ class QuicCongestionControl(abc.ABC):
         self.congestion_window = K_INITIAL_WINDOW * max_datagram_size
 
     @abc.abstractmethod
-    def on_packet_acked(self, *, packet: QuicSentPacket) -> None:
+    def on_packet_acked(self, *, now: float, packet: QuicSentPacket) -> None:
         ...  # pragma: no cover
 
     @abc.abstractmethod
@@ -39,6 +39,12 @@ class QuicCongestionControl(abc.ABC):
     @abc.abstractmethod
     def on_rtt_measurement(self, *, now: float, rtt: float) -> None:
         ...  # pragma: no cover
+
+    def get_log_data(self) -> Dict[str, Any]:
+        data = {"cwnd": self.congestion_window, "bytes_in_flight": self.bytes_in_flight}
+        if self.ssthresh is not None:
+            data["ssthresh"] = self.ssthresh
+        return data
 
 
 class QuicCongestionControlFactory(Protocol):
--- a/src/aioquic/quic/congestion/cubic.py
+++ b/src/aioquic/quic/congestion/cubic.py
@@ -1,0 +1,212 @@
+from typing import Any, Dict, Iterable
+
+from ..packet_builder import QuicSentPacket
+from .base import (
+    K_INITIAL_WINDOW,
+    K_MINIMUM_WINDOW,
+    QuicCongestionControl,
+    QuicRttMonitor,
+    register_congestion_control,
+)
+
+# cubic specific variables (see https://www.rfc-editor.org/rfc/rfc9438.html#name-definitions)
+K_CUBIC_C = 0.4
+K_CUBIC_LOSS_REDUCTION_FACTOR = 0.7
+K_CUBIC_MAX_IDLE_TIME = 2  # reset the cwnd after 2 seconds of inactivity
+
+
+def better_cube_root(x: float) -> float:
+    if x < 0:
+        # avoid precision errors that make the cube root returns an imaginary number
+        return -((-x) ** (1.0 / 3.0))
+    else:
+        return (x) ** (1.0 / 3.0)
+
+
+class CubicCongestionControl(QuicCongestionControl):
+    """
+    Cubic congestion control implementation for aioquic
+    """
+
+    def __init__(self, max_datagram_size: int) -> None:
+        super().__init__(max_datagram_size=max_datagram_size)
+        # increase by one segment
+        self.additive_increase_factor: int = max_datagram_size
+        self._max_datagram_size: int = max_datagram_size
+        self._congestion_recovery_start_time = 0.0
+
+        self._rtt_monitor = QuicRttMonitor()
+
+        self.rtt = 0.02  # starting RTT is considered to be 20ms
+
+        self.reset()
+
+        self.last_ack = 0.0
+
+    def W_cubic(self, t) -> int:
+        W_max_segments = self._W_max / self._max_datagram_size
+        target_segments = K_CUBIC_C * (t - self.K) ** 3 + (W_max_segments)
+        return int(target_segments * self._max_datagram_size)
+
+    def is_reno_friendly(self, t) -> bool:
+        return self.W_cubic(t) < self._W_est
+
+    def is_concave(self) -> bool:
+        return self.congestion_window < self._W_max
+
+    def reset(self) -> None:
+        self.congestion_window = K_INITIAL_WINDOW * self._max_datagram_size
+        self.ssthresh = None
+
+        self._first_slow_start = True
+        self._starting_congestion_avoidance = False
+        self.K: float = 0.0
+        self._W_est = 0
+        self._cwnd_epoch = 0
+        self._t_epoch = 0.0
+        self._W_max = self.congestion_window
+
+    def on_packet_acked(self, *, now: float, packet: QuicSentPacket) -> None:
+        self.bytes_in_flight -= packet.sent_bytes
+        self.last_ack = packet.sent_time
+
+        if self.ssthresh is None or self.congestion_window < self.ssthresh:
+            # slow start
+            self.congestion_window += packet.sent_bytes
+        else:
+            # congestion avoidance
+            if self._first_slow_start and not self._starting_congestion_avoidance:
+                # exiting slow start without having a loss
+                self._first_slow_start = False
+                self._W_max = self.congestion_window
+                self._t_epoch = now
+                self._cwnd_epoch = self.congestion_window
+                self._W_est = self._cwnd_epoch
+                # calculate K
+                W_max_segments = self._W_max / self._max_datagram_size
+                cwnd_epoch_segments = self._cwnd_epoch / self._max_datagram_size
+                self.K = better_cube_root(
+                    (W_max_segments - cwnd_epoch_segments) / K_CUBIC_C
+                )
+
+            # initialize the variables used at start of congestion avoidance
+            if self._starting_congestion_avoidance:
+                self._starting_congestion_avoidance = False
+                self._first_slow_start = False
+                self._t_epoch = now
+                self._cwnd_epoch = self.congestion_window
+                self._W_est = self._cwnd_epoch
+                # calculate K
+                W_max_segments = self._W_max / self._max_datagram_size
+                cwnd_epoch_segments = self._cwnd_epoch / self._max_datagram_size
+                self.K = better_cube_root(
+                    (W_max_segments - cwnd_epoch_segments) / K_CUBIC_C
+                )
+
+            self._W_est = int(
+                self._W_est
+                + self.additive_increase_factor
+                * (packet.sent_bytes / self.congestion_window)
+            )
+
+            t = now - self._t_epoch
+
+            target: int = 0
+            W_cubic = self.W_cubic(t + self.rtt)
+            if W_cubic < self.congestion_window:
+                target = self.congestion_window
+            elif W_cubic > 1.5 * self.congestion_window:
+                target = int(self.congestion_window * 1.5)
+            else:
+                target = W_cubic
+
+            if self.is_reno_friendly(t):
+                # reno friendly region of cubic
+                # (https://www.rfc-editor.org/rfc/rfc9438.html#name-reno-friendly-region)
+                self.congestion_window = self._W_est
+            elif self.is_concave():
+                # concave region of cubic
+                # (https://www.rfc-editor.org/rfc/rfc9438.html#name-concave-region)
+                self.congestion_window = int(
+                    self.congestion_window
+                    + (
+                        (target - self.congestion_window)
+                        * (self._max_datagram_size / self.congestion_window)
+                    )
+                )
+            else:
+                # convex region of cubic
+                # (https://www.rfc-editor.org/rfc/rfc9438.html#name-convex-region)
+                self.congestion_window = int(
+                    self.congestion_window
+                    + (
+                        (target - self.congestion_window)
+                        * (self._max_datagram_size / self.congestion_window)
+                    )
+                )
+
+    def on_packet_sent(self, *, packet: QuicSentPacket) -> None:
+        self.bytes_in_flight += packet.sent_bytes
+        if self.last_ack == 0.0:
+            return
+        elapsed_idle = packet.sent_time - self.last_ack
+        if elapsed_idle >= K_CUBIC_MAX_IDLE_TIME:
+            self.reset()
+
+    def on_packets_expired(self, *, packets: Iterable[QuicSentPacket]) -> None:
+        for packet in packets:
+            self.bytes_in_flight -= packet.sent_bytes
+
+    def on_packets_lost(self, *, now: float, packets: Iterable[QuicSentPacket]) -> None:
+        lost_largest_time = 0.0
+        for packet in packets:
+            self.bytes_in_flight -= packet.sent_bytes
+            lost_largest_time = packet.sent_time
+
+        # start a new congestion event if packet was sent after the
+        # start of the previous congestion recovery period.
+        if lost_largest_time > self._congestion_recovery_start_time:
+            self._congestion_recovery_start_time = now
+
+            # Normal congestion handle, can't be used in same time as fast convergence
+            # self._W_max = self.congestion_window
+
+            # fast convergence
+            if self._W_max is not None and self.congestion_window < self._W_max:
+                self._W_max = int(
+                    self.congestion_window * (1 + K_CUBIC_LOSS_REDUCTION_FACTOR) / 2
+                )
+            else:
+                self._W_max = self.congestion_window
+
+            # normal congestion MD
+            flight_size = self.bytes_in_flight
+            new_ssthresh = max(
+                int(flight_size * K_CUBIC_LOSS_REDUCTION_FACTOR),
+                K_MINIMUM_WINDOW * self._max_datagram_size,
+            )
+            self.ssthresh = new_ssthresh
+            self.congestion_window = max(
+                self.ssthresh, K_MINIMUM_WINDOW * self._max_datagram_size
+            )
+
+            # restart a new congestion avoidance phase
+            self._starting_congestion_avoidance = True
+
+    def on_rtt_measurement(self, *, now: float, rtt: float) -> None:
+        self.rtt = rtt
+        # check whether we should exit slow start
+        if self.ssthresh is None and self._rtt_monitor.is_rtt_increasing(
+            rtt=rtt, now=now
+        ):
+            self.ssthresh = self.congestion_window
+
+    def get_log_data(self) -> Dict[str, Any]:
+        data = super().get_log_data()
+
+        data["cubic-wmax"] = int(self._W_max)
+
+        return data
+
+
+register_congestion_control("cubic", CubicCongestionControl)
--- a/src/aioquic/quic/congestion/reno.py
+++ b/src/aioquic/quic/congestion/reno.py
@@ -23,7 +23,7 @@ class RenoCongestionControl(QuicCongestionControl):
         self._congestion_stash = 0
         self._rtt_monitor = QuicRttMonitor()
 
-    def on_packet_acked(self, *, packet: QuicSentPacket) -> None:
+    def on_packet_acked(self, *, now: float, packet: QuicSentPacket) -> None:
         self.bytes_in_flight -= packet.sent_bytes
 
         # don't increase window in congestion recovery
--- a/src/aioquic/quic/recovery.py
+++ b/src/aioquic/quic/recovery.py
@@ -2,7 +2,7 @@ import logging
 import math
 from typing import Any, Callable, Dict, Iterable, List, Optional
 
-from .congestion import reno  # noqa
+from .congestion import cubic, reno  # noqa
 from .congestion.base import K_GRANULARITY, create_congestion_control
 from .logger import QuicLoggerTrace
 from .packet_builder import QuicDeliveryState, QuicSentPacket
@@ -199,7 +199,7 @@ class QuicPacketRecovery:
                     is_ack_eliciting = True
                     space.ack_eliciting_in_flight -= 1
                 if packet.in_flight:
-                    self._cc.on_packet_acked(packet=packet)
+                    self._cc.on_packet_acked(packet=packet, now=now)
                 largest_newly_acked = packet_number
                 largest_sent_time = packet.sent_time
 
@@ -334,12 +334,7 @@ class QuicPacketRecovery:
         return loss_space
 
     def _log_metrics_updated(self, log_rtt=False) -> None:
-        data: Dict[str, Any] = {
-            "bytes_in_flight": self._cc.bytes_in_flight,
-            "cwnd": self._cc.congestion_window,
-        }
-        if self._cc.ssthresh is not None:
-            data["ssthresh"] = self._cc.ssthresh
+        data: Dict[str, Any] = self._cc.get_log_data()
 
         if log_rtt:
             data.update(
--- a/tests/test_recovery_cubic.py
+++ b/tests/test_recovery_cubic.py
@@ -1,0 +1,584 @@
+import math
+from unittest import TestCase
+
+from aioquic import tls
+from aioquic.quic.congestion.base import K_INITIAL_WINDOW, K_MINIMUM_WINDOW
+from aioquic.quic.congestion.cubic import (
+    K_CUBIC_C,
+    K_CUBIC_LOSS_REDUCTION_FACTOR,
+    CubicCongestionControl,
+    better_cube_root,
+)
+from aioquic.quic.packet import PACKET_TYPE_INITIAL, PACKET_TYPE_ONE_RTT
+from aioquic.quic.packet_builder import QuicSentPacket
+from aioquic.quic.rangeset import RangeSet
+from aioquic.quic.recovery import QuicPacketRecovery, QuicPacketSpace
+
+
+def send_probe():
+    pass
+
+
+def W_cubic(t, K, W_max):
+    return K_CUBIC_C * (t - K) ** 3 + (W_max)
+
+
+class QuicPacketRecoveryCubicTest(TestCase):
+    def setUp(self):
+        self.INITIAL_SPACE = QuicPacketSpace()
+        self.HANDSHAKE_SPACE = QuicPacketSpace()
+        self.ONE_RTT_SPACE = QuicPacketSpace()
+
+        self.recovery = QuicPacketRecovery(
+            congestion_control_algorithm="cubic",
+            initial_rtt=0.1,
+            max_datagram_size=1280,
+            peer_completed_address_validation=True,
+            send_probe=send_probe,
+        )
+        self.recovery.spaces = [
+            self.INITIAL_SPACE,
+            self.HANDSHAKE_SPACE,
+            self.ONE_RTT_SPACE,
+        ]
+
+    def test_better_cube_root(self):
+        self.assertAlmostEqual(better_cube_root(8), 2)
+        self.assertAlmostEqual(better_cube_root(-8), -2)
+        self.assertAlmostEqual(better_cube_root(0), 0)
+        self.assertAlmostEqual(better_cube_root(27), 3)
+
+    def test_discard_space(self):
+        self.recovery.discard_space(self.INITIAL_SPACE)
+
+    def test_on_ack_received_ack_eliciting(self):
+        packet = QuicSentPacket(
+            epoch=tls.Epoch.ONE_RTT,
+            in_flight=True,
+            is_ack_eliciting=True,
+            is_crypto_packet=False,
+            packet_number=0,
+            packet_type=PACKET_TYPE_ONE_RTT,
+            sent_bytes=1280,
+            sent_time=0.0,
+        )
+        space = self.ONE_RTT_SPACE
+
+        #  packet sent
+        self.recovery.on_packet_sent(packet=packet, space=space)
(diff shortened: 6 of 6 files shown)


src/aioquic/quic/configuration.py
```diff
@@ -1,7 +1,8 @@
 from dataclasses import dataclass, field
 from os import PathLike
 from re import split
-from typing import Any, List, Optional, TextIO, Union
+from typing import Any, List, Optional, TextIO, Union, Dict
```
jlaine, line 4: `Dict` is not used

src/aioquic/quic/logger.py
```diff
@@ -32,6 +32,20 @@
 def hexdump(data: bytes) -> str:
     return binascii.hexlify(data).decode("ascii")
 
+def is_builtin_class(Object):
+    return Object.__class__.__module__ == 'builtins'
+
+def get_dataclass_attr(Object) -> dict:
```
jlaine, line 38: Why is this needed?

jlaine:
This PR still seems to include unrelated code such as changes to slowstart, minmax.. can we cut it down to the bare minimum to support Cubic please?

Aperence:
Hi, 

First, thanks for your feedback.

I have normally removed all content that weren't related to cubic (as slow-start, minmax, get_dataclass_attr, ...).

Only the code for cubic should remain now.

Best,

src/aioquic/quic/logger.py
Aperence: It was needed for a simpler logging of the BBR/BBRv2 congestion control, where many attributes were present, it was thus better to have some method that automatically gets all its attributes.

This won't be needed for cubic, so I'll simply remove it for now

src/aioquic/quic/congestion/congestion.py
```diff
@@ -0,0 +1,155 @@
...
+        self._max_datagram_size = max_datagram_size
+        self.cwnd = fixed_cwnd
+        self.data_in_flight = 0
+
+    def set_recovery(self, recovery):
+        # recovery is a QuicPacketRecovery instance
+        self.recovery = recovery
+
+    def on_init(self, *args, **kwargs):
+        pass
```
jlaine, line 93: This is unused AFAICT?

src/aioquic/quic/congestion/cubic.py
```diff
@@ -0,0 +1,222 @@
...
+K_CUBIC_MAX_IDLE_TIME = 2   # reset the cwnd after 2 seconds of inactivity
+
+class CubicCongestionControl(QuicCongestionControl):
+    """
+    Cubic congestion control implementation for aioquic
+    """
+
+    def __init__(self, max_datagram_size : int, callback=None, reno_friendly_activated = True) -> None:
+        super().__init__(max_datagram_size=max_datagram_size, callback=callback)
+        global K_CUBIC_ADDITIVE_INCREASE
```
jlaine, line 19: Uhoh, this doesn't look good, this is de-factor not a constant

src/aioquic/quic/congestion/congestion.py
```diff
@@ -0,0 +1,155 @@
...
+                    self._increases += 1
+                    if self._increases >= self._size:
+                        return True
+                elif delta > 0:
+                    self._increases = 0
+        return False
+    
+
+def Now():
+    return datetime.timestamp(datetime.now())
```
jlaine, line 76: I'm not sure I'm comfortable with this, until now all timestamps have been provided as input to the congestion controller's method. 

If I remember correctly, this timestamp is handed to us by the event loop, and there is no guarantee its aligns with `datetime.now`.

Aperence:
Hello, 

I modified the content you pinpointed above, feel free to still give me feedback is anything is still weird or should be changed :D

Have a nice day,

examples/http3_client.py
```diff
@@ -545,6 +550,14 @@ async def main(
...
     if args.secrets_log:
         configuration.secrets_log_file = open(args.secrets_log, "a")
+    if args.congestion_control:
+        if args.congestion_control == "cubic":
+            configuration.congestion_control = CubicCongestionControl()
+        elif args.congestion_control == "reno":
+            configuration.congestion_control = RenoCongestionControl()
+        else:
+            print("Invalid congestion control algorithm")
+            exit(127)
```
jlaine, line 560:
```suggestion
            raise Exception("Invalid congestion control algorithm")
```

src/aioquic/quic/congestion/__init__.py
```diff
@@ -0,0 +1 @@
+from .congestion import *
```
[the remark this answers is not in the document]
Aperence: This feels indeed a better way to do it

jlaine:
OK we're getting close to a mergeable state! Would you mind rebasing on top of `main` and squashing down all your commits so we can get a CI run?

Aperence:
Normally this should be done, I squashed all the commits into one (4cf275a34e0c8fa30c29b9e9441f0adb31e4015a)

src/aioquic/quic/congestion/__init__.py
```diff
@@ -0,0 +1,145 @@
...
+    def on_rtt_measurement(self, latest_rtt: float, now: float) -> None:
+        if self.callback:
+            self.callback(CongestionEvent.RTT_MEASURED, self)
+
+    def get_congestion_window(self) -> int:
+        # return the cwnd in number of bytes
+        return self.cwnd   
+    
+    def _set_congestion_window(self, value):
+        self.cwnd = value
```
jlaine, line 126: I'm having trouble understanding why we have these accessors?

src/aioquic/quic/congestion/__init__.py
```diff
@@ -0,0 +1,145 @@
...
+                # don't call this if it is a superclass that runs
+                self.data_in_flight -= packet.sent_bytes
+
+    def on_packets_lost(self, packets: Iterable[QuicSentPacket], now: float) -> None:
+        if self.callback:
+            self.callback(CongestionEvent.PACKET_LOST, self)
+        if type(self) == QuicCongestionControl:
+            for packet in packets:
+                # don't call this if it is a superclass that runs
+                self.data_in_flight -= packet.sent_bytes
```
jlaine, line 115: This looks very odd..

src/aioquic/quic/congestion/__init__.py
```diff
@@ -0,0 +1,145 @@
...
+        return self.cwnd   
+    
+    def _set_congestion_window(self, value):
+        self.cwnd = value
+
+    def get_ssthresh(self) -> Optional[int]: 
+        pass
+
+    def get_bytes_in_flight(self) -> int:
+        return self.data_in_flight
```
jlaine, line 132: Same as above, why do we need this accessor, especially since `data_in_flight` doesn't seem to be set anywhere other than at initialisation?

src/aioquic/quic/congestion/__init__.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def log_callback(self) -> Dict[str, Any]:
```
jlaine, line 134: This doesn't look like it's actually a "callback", it looks more like `get_log_data` or something?

src/aioquic/quic/congestion/__init__.py
Aperence: It is used for having an uniform access to the congestion window, as the variable names might change according to the implementation and the developer, thus by providing a method that must be implemented on each class extending QuicCongestionControll, we ensure that whatever the variable name is, we can still access the congestion window (Note however that the setter is unused, so I'll just remove it)

src/aioquic/quic/congestion/__init__.py
Aperence: It is an artifact of how bytes in flight was used, I'll refactor this

src/aioquic/quic/congestion/__init__.py
Aperence: Same as above, to provide an uniform interface for getting the bytes in flight, this is needed by `recovery.py` at line 134, which is used typically in `connection.py`, line 580 (this was already there before, just changed it to create a getter)

src/aioquic/quic/congestion/__init__.py
Aperence: It is indeed used to get the logging data for the congestion controller, thus I renamed it

jlaine:
I've been reading more of the code and I'm not entirely comfortable with how it's architectured. Could you take a look at PR #432 which *only* moves code around and introduces a base class for congestion control algorithms?

rmarx:
Hey both,

(just for the QLOG stuff ;))

qlog is actually intended to be extensible, so implementations can log whatever extra data they care about. We indeed don't want to over-specify for the different CC algorithms, but other implementations have certainly also logged their own fields for that. The one thing you can't count on of course is that general tools (like QVIS) will do anything more with your custom fields than just show them (which is what qvis does). 

Looking at the specific fields here, they seem sensible. My suggestion would be to start the fields with `cubic-` though, and make them lowercase (that IS required by qlog ;)), so `cubic-phase` or `cubic-wmax`.

src/aioquic/quic/congestion/cubic.py
```diff
@@ -0,0 +1,210 @@
...
+
+
+    def on_rtt_measurement(self, latest_rtt: float, now: float) -> None:
+        self.now = now
+        super().on_rtt_measurement(latest_rtt, now)
+        # check whether we should exit slow start
+        if self.ssthresh is None and self._rtt_monitor.is_rtt_increasing(
+            latest_rtt, now
+        ):
+            self._cwnd_prior = self.congestion_window
```
jlaine, line 181: This code does nothing, `self._cwnd_prior` is never used anywhere?

src/aioquic/quic/congestion/cubic.py
```diff
@@ -0,0 +1,210 @@
...
+        self._starting_congestion_avoidance = False
+        self.K = 0
+        self._W_est = 0
+        self._cwnd_epoch = 0
+        self._t_epoch = 0
+        self._W_max = self.congestion_window
+
+    def on_packet_acked(self, packet: QuicSentPacket, now: float) -> None:
+        super().on_packet_acked(packet, now)
+        rtt = self.recovery._rtt_smoothed
```
jlaine, line 72: No, the congestion control should not have a handle to the `QuicPacketRecovery`. Is it actually the smooth RTT we want here or the latest?

src/aioquic/quic/congestion/cubic.py
```diff
@@ -0,0 +1,210 @@
...
+        self._cwnd_epoch = None
+        self._t_epoch = None
+        self._W_max = None
+        self._first_slow_start = True
+        self._starting_congestion_avoidance = False
+        self.K = 0
+        self._W_est = 0
+        self._cwnd_epoch = 0
+        self._t_epoch = 0
+        self._W_max = self.congestion_window
```
jlaine, line 68: There are multiple redundant initialisations of the same variables..

src/aioquic/quic/congestion/cubic.py
```diff
@@ -0,0 +1,210 @@
...
+        return target_segments * self._max_datagram_size
+    
+    def is_reno_friendly(self, t) -> bool:
+        return self.reno_friendly_activated and self.W_cubic(t) < self._W_est
+    
+    def is_concave(self):
+        return self.congestion_window < self._W_max
+    
+    def is_convex(self):
+        return self.congestion_window >= self._W_max
```
jlaine, line 53: This is not used, it's just a negation of `is_concave`

src/aioquic/quic/congestion/cubic.py
Aperence: It should be according to the RFC the weighted average RTT

src/aioquic/quic/congestion/cubic.py
Aperence: Fixed

src/aioquic/quic/congestion/cubic.py
Aperence: But I'm wondering if we couldn't use the latest RTT indeed ?

examples/http3_client.py
```diff
@@ -518,6 +518,9 @@ async def main(
     parser.add_argument(
         "--zero-rtt", action="store_true", help="try to send requests using 0-RTT"
     )
+    parser.add_argument(
+        "--congestion-control", type=str, help="which congestion control algorithm to use (reno, cubic)"
+    )
```
jlaine, line 523: Not needed, already in `main`

examples/http3_server.py
```diff
@@ -564,6 +564,9 @@ async def main(
     parser.add_argument(
         "-v", "--verbose", action="store_true", help="increase logging verbosity"
     )
+    parser.add_argument(
+        "--congestion-control", type=str, help="which congestion control algorithm to use (reno, cubic)"
+    )
```
jlaine, line 569: Not needed, already in `main`

src/aioquic/quic/congestion/base.py
```diff
@@ -17,15 +17,16 @@ class QuicCongestionControl(abc.ABC):
     congestion_window: int = 0
     ssthresh: Optional[int] = None
 
-    def __init__(self, *, max_datagram_size: int) -> None:
+    def __init__(self, *, max_datagram_size: int, recovery) -> None:
         self.congestion_window = K_INITIAL_WINDOW * max_datagram_size
+        self.recovery = recovery
```
jlaine, line 22: No, let's not have a reference loop please

src/aioquic/quic/congestion/base.py
```diff
@@ -17,15 +17,16 @@ class QuicCongestionControl(abc.ABC):
...
 
     @abc.abstractmethod
-    def on_packet_acked(self, *, packet: QuicSentPacket) -> None:
+    def on_packet_acked(self, *, now : float, packet: QuicSentPacket) -> None:
```
jlaine, line 25: Why do you need `now`, you already have `packet.sent_time`?

src/aioquic/quic/congestion/base.py
```diff
@@ -17,15 +17,16 @@ class QuicCongestionControl(abc.ABC):
...
         ...  # pragma: no cover
 
     @abc.abstractmethod
-    def on_packet_sent(self, *, packet: QuicSentPacket) -> None:
+    def on_packet_sent(self, *, now : float, packet: QuicSentPacket) -> None:
```
jlaine, line 29: Why do you need `now`, you already have `packet.sent_time`?

src/aioquic/quic/congestion/cubic.py
```diff
@@ -0,0 +1,204 @@
...
+        self.last_ack = None
+
+        self.now = 0
+
+    def better_cube_root(self, x):
+        if (x < 0):
+            # avoid precision errors that make the cube root returns an imaginary number
+            return -((-x)**(1./3.))
+        else:
+            return (x)**(1./3.)
```
jlaine, line 42: Move this to a standalone function, it will avoid re-defining it in `test_cubic.py`

src/aioquic/quic/congestion/cubic.py
jlaine: I don't know the RFCs enough, if you need it add a `smoothed_rtt` keyword argument to `on_rtt_measurement`

src/aioquic/quic/congestion/base.py
Aperence: How should I get the smoothed RTT in another way ?

src/aioquic/quic/recovery.py
```diff
@@ -62,7 +63,10 @@ def update_bucket(self, now: float) -> None:
             self.evaluation_time = now
 
     def update_rate(self, congestion_window: int, smoothed_rtt: float) -> None:
-        pacing_rate = congestion_window / max(smoothed_rtt, K_MICRO_SECOND)
+        if self.pacing_rate == None or self.pacing_rate == 0:
+            pacing_rate = congestion_window / max(smoothed_rtt, K_MICRO_SECOND)
+        else:
+            pacing_rate = self.pacing_rate
```
jlaine, line 69: Why are we changing this?

src/aioquic/quic/recovery.py
```diff
@@ -115,17 +119,17 @@ def __init__(
...
         self._cc = create_congestion_control(
-            congestion_control_algorithm, max_datagram_size=max_datagram_size
+            congestion_control_algorithm, max_datagram_size=max_datagram_size, recovery=self
         )
         self._pacer = QuicPacketPacer(max_datagram_size=max_datagram_size)
 
     @property
     def bytes_in_flight(self) -> int:
-        return self._cc.bytes_in_flight
+        return int(self._cc.bytes_in_flight)
```
jlaine, line 128: This is already an int, no need to cast it.

src/aioquic/quic/congestion/base.py
Aperence: Changed to latest RTT instead

src/aioquic/quic/congestion/base.py
```diff
@@ -1,5 +1,5 @@
 import abc
-from typing import Dict, Iterable, Optional, Protocol
+from typing import Dict, Iterable, Optional, Protocol, Any
```
jlaine, line 2: Please run the linters (black / ruff / mypy) locally to avoid CI runs,  this is guaranteed to not be OK as it's not sorted

src/aioquic/quic/congestion/cubic.py
```diff
@@ -0,0 +1,206 @@
...
+            flight_size = self.bytes_in_flight
+            new_ssthresh = max(int(flight_size*K_CUBIC_LOSS_REDUCTION_FACTOR), K_MINIMUM_WINDOW * self._max_datagram_size)
+            self.ssthresh = new_ssthresh
+            self.congestion_window = max(self.ssthresh, K_MINIMUM_WINDOW * self._max_datagram_size)
+            
+
+            self._starting_congestion_avoidance = True  # restart a new congestion avoidance phase
+
+
+    def on_rtt_measurement(self, latest_rtt: float, now: float) -> None:
```
jlaine, line 172: `rtt` not `latest_rtt`

src/aioquic/quic/congestion/cubic.py
```diff
@@ -0,0 +1,206 @@
...
+
+        self._rtt_monitor = QuicRttMonitor()
+
+        self.rtt = 0.02    # starting RTT is considered to be 20ms
+
+        self.reset()
+
+        self.last_ack = None
+
+        self.now = 0
```
jlaine, line 44: Please remove this, the controller cannot know the "current time" as it has no clock. From what I can see the only place you use `self.now` is in `get_log_data`, you can pass it in there.

src/aioquic/quic/recovery.py
Aperence: Woopsie, seems like an artifact of BBR (where the congestion controller needed to set the pacing rate), just removed it

jlaine:
Please revert your changes to `reno.py`, the only change should be the addition of `now` in `on_packet_acked`.

src/aioquic/quic/connection.py
```diff
@@ -3318,4 +3319,4 @@ def _write_streams_blocked_frame(
                     is_unidirectional=frame_type == QuicFrameType.STREAMS_BLOCKED_UNI,
                     limit=limit,
                 )
-            )
+            )
```
jlaine, line 3322: Please revert all changes to `connection.py`, they are not needed

src/aioquic/quic/logger.py
```diff
@@ -32,7 +32,6 @@
 def hexdump(data: bytes) -> str:
     return binascii.hexlify(data).decode("ascii")
 
-
```
jlaine, line 35: Please revert all changes to `logger.py`, they are not needed

src/aioquic/quic/recovery.py
```diff
@@ -1,8 +1,9 @@
 import logging
 import math
 from typing import Any, Callable, Dict, Iterable, List, Optional
+from datetime import datetime
```
jlaine, line 4: This is sure to fail linting, use `black` and `ruff --fix` locally to fix everything in `src`

src/aioquic/quic/recovery.py
```diff
@@ -238,7 +238,7 @@ def on_ack_received(
                 )
 
             # inform congestion controller
-            self._cc.on_rtt_measurement(now=now, rtt=latest_rtt)
+            self._cc.on_rtt_measurement(now=now, latest_rtt=latest_rtt)
```
jlaine, line 241: Revert this

src/aioquic/quic/recovery.py
```diff
@@ -263,7 +263,7 @@ def on_loss_detection_timeout(self, *, now: float) -> None:
             self._pto_count += 1
             self.reschedule_data(now=now)
 
-    def on_packet_sent(self, *, packet: QuicSentPacket, space: QuicPacketSpace) -> None:
+    def on_packet_sent(self, packet: QuicSentPacket, space: QuicPacketSpace) -> None:
```
jlaine, line 266: Revert this

tests/test_recovery.py
```diff
@@ -1,7 +1,20 @@
 from unittest import TestCase
 
+<<<<<<< HEAD
```
jlaine, line 3: Revert all these changes

src/aioquic/quic/congestion/cubic.py
```diff
@@ -0,0 +1,206 @@
...
+from ..packet_builder import QuicSentPacket
+from typing import Iterable, Dict, Any
+
+# cubic specific variables (see https://www.rfc-editor.org/rfc/rfc9438.html#name-definitions)
+K_CUBIC_K = 1    
+K_CUBIC_C = 0.4
+K_CUBIC_LOSS_REDUCTION_FACTOR = 0.7
+K_CUBIC_MAX_IDLE_TIME = 2   # reset the cwnd after 2 seconds of inactivity
+
+def better_cube_root(self, x):
```
jlaine, line 17: self? While we are at it, please add type annotations otherwise `mypy` will fail

src/aioquic/quic/congestion/reno.py
```diff
@@ -41,14 +41,14 @@ def on_packet_acked(self, *, packet: QuicSentPacket) -> None:
                 self._congestion_stash -= count * self.congestion_window
                 self.congestion_window += count * self._max_datagram_size
 
-    def on_packet_sent(self, *, packet: QuicSentPacket) -> None:
+    def on_packet_sent(self, packet: QuicSentPacket) -> None:
```
jlaine, line 44: The switch to keyword-only arguments was on purpose, revert this

src/aioquic/quic/congestion/reno.py
```diff
@@ -41,14 +41,14 @@ def on_packet_acked(self, *, packet: QuicSentPacket) -> None:
...
         self.bytes_in_flight += packet.sent_bytes
 
-    def on_packets_expired(self, *, packets: Iterable[QuicSentPacket]) -> None:
+    def on_packets_expired(self, packets: Iterable[QuicSentPacket]) -> None:
```
jlaine, line 47: Ditto

src/aioquic/quic/congestion/reno.py
```diff
@@ -41,14 +41,14 @@ def on_packet_acked(self, *, packet: QuicSentPacket) -> None:
...
         for packet in packets:
             self.bytes_in_flight -= packet.sent_bytes
 
-    def on_packets_lost(self, *, now: float, packets: Iterable[QuicSentPacket]) -> None:
+    def on_packets_lost(self, packets: Iterable[QuicSentPacket], now: float) -> None:
```
jlaine, line 51: Ditto

src/aioquic/quic/congestion/cubic.py
```diff
@@ -0,0 +1,206 @@
...
+    K_INITIAL_WINDOW,
+    QuicCongestionControl,
+    QuicRttMonitor,
+    register_congestion_control,
+)
+from ..packet_builder import QuicSentPacket
+from typing import Iterable, Dict, Any
+
+# cubic specific variables (see https://www.rfc-editor.org/rfc/rfc9438.html#name-definitions)
+K_CUBIC_K = 1
```
jlaine, line 12: Where is `K_CUBIC_K` used?

src/aioquic/quic/recovery.py
jlaine: > In cubic, this can be a float, leading to some unexpected error, so I just casted it when trying to access it

Uh no I don't see how this can be a float, we start with 0 and all we do is add / subtract packet lengths. This cannot and must not be a float.

src/aioquic/quic/congestion/cubic.py
```diff
@@ -0,0 +1,206 @@
...
+
+            t = now - self._t_epoch
+            
+            target = None
+            if (self.W_cubic(t + self.rtt) < self.congestion_window):
+                target = self.congestion_window
+            elif (self.W_cubic(t + self.rtt) > 1.5*self.congestion_window):
+                target = self.congestion_window*1.5
+            else:
+                target = self.W_cubic(t + self.rtt)
```
jlaine, line 114: No need to re-calculate this 3 times, just assign `target = self.W_cubic(t + self.rtt)` then clamp it

src/aioquic/quic/congestion/reno.py
```diff
@@ -66,12 +66,22 @@ def on_packets_lost(self, *, now: float, packets: Iterable[QuicSentPacket]) -> N
 
         # TODO : collapse congestion window if persistent congestion
 
-    def on_rtt_measurement(self, *, now: float, rtt: float) -> None:
+    def on_rtt_measurement(self, latest_rtt: float, now: float) -> None:
         # check whether we should exit slow start
         if self.ssthresh is None and self._rtt_monitor.is_rtt_increasing(
-            now=now, rtt=rtt
+            rtt=latest_rtt, now=now
```
jlaine, line 72: Revert this too

src/aioquic/quic/congestion/reno.py
```diff
@@ -66,12 +66,22 @@ def on_packets_lost(self, *, now: float, packets: Iterable[QuicSentPacket]) -> N
...
         ):
             self.ssthresh = self.congestion_window
-
+    
+    def get_log_data(self) -> Dict[str, Any]:
+        data = {
+            "cwnd" : self.congestion_window
+        }
+        if (self.ssthresh == None):
```
jlaine, line 80:
```suggestion
        if self.ssthresh is None:
```

Aperence:
Ok, I've reverted normally every changed except for cubic, and logs, I've also run the lint checks locally (ruff + black + mypy, thank you btw because I didn't know about these tools at all) and I fixed normally everything that was listed locally

Aperence:
My black version locally seem however to differ compared to the one used on CI, as it fails here, I can fix this (removing a blank line), if needed, but I'm waiting for you to confirm as I might have abused you CI instances with all these push, my apologies for this :cry:

jlaine:
It looks as though you still have one linter error left :)

The code is also going to need thorough tests to check everything is working in all scenarios. I would suggest you try writing a `test_recovery_cubic.py` based on `test_recovery_reno.py`. Full test coverage is a requirement to merge :)

src/aioquic/quic/congestion/base.py
```diff
@@ -40,6 +40,12 @@ def on_packets_lost(self, *, now: float, packets: Iterable[QuicSentPacket]) -> N
     def on_rtt_measurement(self, *, now: float, rtt: float) -> None:
         ...  # pragma: no cover
 
+    def get_log_data(self) -> Dict[str, Any]:
+        data = {"cwnd": self.congestion_window}
```
jlaine, line 44: Looks like you dropped `bytes_in_flight` here

src/aioquic/quic/recovery.py
```diff
@@ -334,10 +334,7 @@ def _get_loss_space(self) -> Optional[QuicPacketSpace]:
         return loss_space
 
     def _log_metrics_updated(self, log_rtt=False) -> None:
-        data: Dict[str, Any] = {
-            "bytes_in_flight": self._cc.bytes_in_flight,
-            "cwnd": self._cc.congestion_window,
-        }
+        data: Dict[str, Any] = self._cc.get_log_data()
         if self._cc.ssthresh is not None:
```
jlaine, line 338: This doesn't look needed, we're already reporting `ssthresh` in `QuicCongestionControl.get_log_data()`

src/aioquic/quic/congestion/cubic.py
```diff
@@ -0,0 +1,216 @@
...
+    def is_concave(self) -> bool:
+        return self.congestion_window < self._W_max
+
+    def reset(self) -> None:
+        self.congestion_window = K_INITIAL_WINDOW * self._max_datagram_size
+        self.ssthresh = None
+
+        self._first_slow_start = True
+        self._starting_congestion_avoidance = False
+        self.K: float = 0.0
```
jlaine, line 63: If you initialise K with a float you don't need the type definition

Aperence:
Normally, I've fixed these last issues + added a test for cubic based on `test_recovery_cubic.py`. I'll try to check later the code that wasn't covered by thos tests

jlaine:
> Normally, I've fixed these last issues + added a test for cubic based on `test_recovery_cubic.py`. I'll try to check later the code that wasn't covered by thos tests

Sure! You can get coverage reports locally too:

```shell
pip install coverage
coverage run -m unittest tests/test_recovery_cubic.py
coverage html
firefox htmlcov/index.html
```

jlaine:
Wow, amazing work, thanks so much for your patience! Is the `test_cubic.py` still useful or is it superseded by your `test_recovery_cubic.py`? If it's still useful maybe merge it into `test_recovery_cubic.py`?

Disclaimer: I am not able to determine whether the code works as intended, or follows specs. Would you mind taking the code in its current form for a spin and checking it still behaves as you want it to?

Once this is done I'd be happy to merge the code!

Aperence:
Hi, 

First, thank you for all your feedback during this PR, I learned quite a lot about CI tools to ensure some code quality !
Secondly, about the `test_cubic.py`, it is still useful, so I'll just merge it in `test_recovery_cubic.py` as you recommended, this should centralize all the tests for the cubic congestion controller.

The implementation seems to still work fine, here are some examples on how it performs:

With no delay on localhost:
image

With a delay of 100ms on localhost:
image

We can clearly still see the cubic growth when the delay is quite important (second plot), and the reno friendly region when we have a reasonable delay (first plot), thus I think we can affirm that it still work as intended

jlaine:
Alright, thanks for the checks! Are these graphs produced from the QLOG output? I saw an article which mentioned you had set up a test bench as part of your University work?

If wouldn't mind squashing your commits down to a single commit with a clean commit message I'd be happy to hit the "merge" button.

Aperence:
Yup, these graphs are generated based on the QLOG files, the script can be found [here](https://github.com/Aperence/linfo1341-congestion-student/blob/main/src/plot_congestion.py) if you are interested with it (it support normally any congestion control, as it only check the presence of the cwnd field in the logs), but some features (W_max, ...) might not be up to date with the latest modifications for the logs (but the core for plotting losses and cwnd does). You can also select a particular field if you want to plot other things (for example ssthresh, bytes_in_flight, ...)

Normally, the commits should be squashed also !

jlaine:
Merged, thanks a lot! 🎉

I hope we will have the pleasure of seing more of your work :)

PR #432, opened 2023-12-13
Move congestion control code to allow other algorithms

jlaine:
Use keyword arguments wherever possible to make it explicit what is being passed around.

### DIFF
--- a/examples/http3_client.py
+++ b/examples/http3_client.py
@@ -446,6 +446,12 @@ if __name__ == "__main__":
         ),
     )
     parser.add_argument(
+        "--congestion-control-algorithm",
+        type=str,
+        default="reno",
+        help="use the specified congestion control algorithm",
+    )
+    parser.add_argument(
         "-d", "--data", type=str, help="send the specified data in a POST request"
     )
     parser.add_argument(
@@ -527,6 +533,7 @@ if __name__ == "__main__":
     configuration = QuicConfiguration(
         is_client=True,
         alpn_protocols=H0_ALPN if args.legacy_http else H3_ALPN,
+        congestion_control_algorithm=args.congestion_control_algorithm,
         max_datagram_size=args.max_datagram_size,
     )
     if args.ca_certs:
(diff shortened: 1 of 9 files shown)


jlaine:
@Aperence this is a proposal so that #415 can be simplified.

One thing which is not yet handled is "how do we specify the congestion controller in the settings". PR #415 is not perfect in this respect, as we definitely don't want to pass around *instances* of the controller at the configuration level. I can think of a few options:

- Pass a string which gives the path to the class, e.g. : `QuicConfiguration(congestion_control="aioquic.quic.congestion.reno.RenoCongestionControl"`).  This is inspired by Django's settings for its various backends (e-mail backends, etc..).
- Pass a callable which returns an instance of a  `QuicCongestionControl`  subclass, e.g. : `QuicConfiguration(congestion_control=RenoCongestionControl)`.
- Have a registry of available congestion control algorithms, keyed off a string. `congestion.register("reno", RenoCongestionControl)`, then `QuicConfiguration(congestion_control="reno")`. One advantage of having a registry would be to ease adding support for this throughout our examples without needing to update them whenever there is a new congestion controller.

@rthalley any opinions?

rthalley:
Dnspython has instances of both the callable / factory and the registry.  I tend to use the callable style for cases where you know at coding time what you want to do.  I use the registry style for things you discover at runtime.  So, for example, `dns.zone.from_file()` has an optional `zone_factory` parameter that lets you specify which `Zone` subclass you want created, and defaults to `Zone` itself.  But dynamic things like DNS RR types or DNSSEC algorithms which are decoded at runtime have registries, which allows you to add new types / algorithms in your application without any changes to the code working with types / algorithms.

I don't love the "path to the class" scheme, as given you can pass a first class callable, I don't see the point in referring to it by name.

rthalley:
But to answer your question, I'm not sure about this case as it's almost on the line.  E.g. if you think it's really uncommon to fool around with it, then the callable way seems fine, but if you thought all sorts of aioquic apps are going to expose this in their own interfaces at runtime, e.g. "--congestion-control <scheme>" then the registry seems better as anything registered would just work for the rest of the ecosystem, you wouldn't have to change code in N app that had "--congestion-control" options.

PR #435, opened 2023-12-27
Fix tls.py assertion issues.

rthalley:
1) Some assertions had side effects and would cause a loss of
   framing if python was run with -O or -OO

2) Some assertions should have been error checks.

### DIFF
--- a/src/aioquic/tls.py
+++ b/src/aioquic/tls.py
@@ -100,6 +100,10 @@ class AlertCertificateExpired(Alert):
     description = AlertDescription.certificate_expired
 
 
+class AlertDecodeError(Alert):
+    description = AlertDescription.decode_error
+
+
 class AlertDecryptError(Alert):
     description = AlertDescription.decrypt_error
 
@@ -330,6 +334,10 @@ class HandshakeType(IntEnum):
     MESSAGE_HASH = 254
 
 
+class NameType(IntEnum):
+    HOST_NAME = 0
+
+
 class PskKeyExchangeMode(IntEnum):
     PSK_KE = 0
     PSK_DHE_KE = 1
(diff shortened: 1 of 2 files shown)


rthalley:
I haven't figured out how to run the tests with -O yet, so consider this a draft.  Everything passes for me without -O, and the problem my test server was having when run with -O went away.  Arguably it would be good to add test coverage for the various kinds of input problems that can now cause TLS alerts, but I haven't figured out how to do those yet.

src/aioquic/tls.py
```diff
@@ -365,7 +369,9 @@ def pull_block(buf: Buffer, capacity: int) -> Generator:
     length = int.from_bytes(buf.pull_bytes(capacity), byteorder="big")
     end = buf.tell() + length
     yield length
-    assert buf.tell() == end
+    if buf.tell() != end:
+        # There was trailing garbage or our parsing was bad.
+        raise AlertDecodeError("extra bytes at the end of a block")
```
jlaine, line 374: I think this should remain an assertion. It means the caller of `pull_block` did not read the right number of bytes within the context manager.

src/aioquic/tls.py
```diff
@@ -551,7 +570,12 @@ def pull_extension() -> None:
                 hello.psk_key_exchange_modes = pull_list(buf, 1, buf.pull_uint8)
             elif extension_type == ExtensionType.SERVER_NAME:
                 with pull_block(buf, 2):
-                    assert buf.pull_uint8() == 0
+                    name_type = buf.pull_uint8()
+                    if name_type != 0:
+                        # We don't know this name_type.
+                        raise AlertIllegalParameter(
+                            f"unknown server name type {name_type}"
+                        )
```
jlaine, line 578: I would suggest adding a `pull_server_name` method to facilitate testing. For symmetry you might also consider adding `push_server_name` too.

jlaine: .. and done

src/aioquic/tls.py
jlaine: No scratch that comment.

PR #436, opened 2023-12-28
Raise a TLS alert if parsing a message causes a BufferReadError

jlaine:
While parsing TLS messages, if a `BufferReadError` occurs, it is a sign we are unable to parse the message because it is malformed. In such an event raise a `AlertDecodeError` exception so that the connection gets shut down with a TLS error.


jlaine:
If `tls.Context.handle_message` raises any subclass of `tls.Alert`, the connection gets shut down properly:

https://github.com/aiortc/aioquic/blob/57722463ff6b6cbe11f056d7f253ed5161f25356/src/aioquic/quic/connection.py#L1588

PR #437, opened 2023-12-29
Remove utcnow() deprecation warning.

rthalley:
Python 3.12 deprecates utcnow(), but we cannot do the recommended thing and replace utcnow() with datetime.now(timezone.utc) as the datetime objects that the cryptography x509 library puts in certificates do not have any timezone info.  If we try to compare them with something that has timezone info we will get an exception:

    TypeError: can't compare offset-naive and offset-aware datetimes

We must continue to make an explictly offset-naive timestamp.

### DIFF
--- a/src/aioquic/tls.py
+++ b/src/aioquic/tls.py
@@ -63,7 +63,7 @@ T = TypeVar("T")
 #
 #     TypeError: can't compare offset-naive and offset-aware datetimes
 #
-# We must continue to make an explictly offset-naive timestamp.
+# We must continue to make an explicitly offset-naive timestamp.
 utcnow = partial(datetime.datetime.now, None)
 
 
--- a/src/aioquic/tls.py
+++ b/src/aioquic/tls.py
@@ -54,8 +54,17 @@ SERVER_CONTEXT_STRING = b"TLS 1.3, server CertificateVerify"
 
 T = TypeVar("T")
 
-# facilitate mocking for the test suite
-utcnow = datetime.datetime.utcnow
+# facilitate time mocking for the test suite
+#
+# Python 3.12 deprecates utcnow(), but we cannot do the recommended thing and replace
+# utcnow() with datetime.now(timezone.utc) as the datetime objects that the cryptography
+# x509 library puts in certificates do not have any timezone info.  If we try to
+# compare them with something that has timezone info we will get an exception:
+#
+#     TypeError: can't compare offset-naive and offset-aware datetimes
+#
+# We must continue to make an explictly offset-naive timestamp.
+utcnow = partial(datetime.datetime.now, None)
 
 
 class AlertDescription(IntEnum):
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -32,14 +32,15 @@ def generate_certificate(*, alternative_names, common_name, hash_algorithm, key)
         [x509.NameAttribute(x509.NameOID.COMMON_NAME, common_name)]
     )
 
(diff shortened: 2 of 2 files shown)


src/aioquic/tls.py
```diff
@@ -54,8 +54,17 @@
...
+#
+# Python 3.12 deprecates utcnow(), but we cannot do the recommended thing and replace
+# utcnow() with datetime.now(timezone.utc) as the datetime objects that the cryptography
+# x509 library puts in certificates do not have any timezone info.  If we try to
+# compare them with something that has timezone info we will get an exception:
+#
+#     TypeError: can't compare offset-naive and offset-aware datetimes
+#
+# We must continue to make an explicitly offset-naive timestamp.
+utcnow = partial(datetime.datetime.now, None)
```
jlaine, line 67: Hm, this doesn't look equivalent, I would have expected something like:

`datetime.datetime.now(datetime.timezone.utc).replace(tzinfo=None)`

jlaine: Looking into this a bit deeper it looks like we are going the wrong way here. We should be moving towards timezone-aware datetimes, as in:

- Making `utcnow = partial(datetime.datetime.now, datetime.timezone.utc)`
- Using `certificate.not_valid_before_utc` and `certificate.not_valid_after_utc`: https://cryptography.io/en/latest/x509/reference/#cryptography.x509.Certificate.not_valid_before_utc
- Requiring `cryptography >= 42.0.0`

tests/utils.py
```diff
@@ -32,14 +32,15 @@ def generate_certificate(*, alternative_names, common_name, hash_algorithm, key)
...
     builder = (
         x509.CertificateBuilder()
         .subject_name(subject)
         .issuer_name(issuer)
         .public_key(key.public_key())
         .serial_number(x509.random_serial_number())
-        .not_valid_before(datetime.datetime.utcnow())
-        .not_valid_after(datetime.datetime.utcnow() + datetime.timedelta(days=10))
+        .not_valid_before(datetime.datetime.now(None))
+        .not_valid_after(datetime.datetime.now(None) + datetime.timedelta(days=10))
```
jlaine, line 43: I think here we should use `tls.utcnow()` for consistency

jlaine:
Unless I'm mistaken, the "right" way to fix this would be to wait for `cryptography` 42 which supports timezone-aware datetimes, see my comments above.

rthalley:
If they are going to fix it, then yes, that is the right way!

jlaine:
> If they are going to fix it, then yes, that is the right way!

We could also have a two-step approach.

## Right now

- make `tests/utils.py` use `aioquic.tls.utcnow`
- replace `utcnow` by:

```python
def utcnow() -> datetime.datetime:
    return datetime.datetime.now(datetime.timezone.utc).replace(tzinfo=None)
```

## When cryptography 42 comes out

Use timezone-aware datetimes everywhere!

It would be good to check that `CertificateBuilder` has also been updated to allow timezone-aware datetimes, as I only checked `x509.Certificate` attributes.

rthalley:
I checked CertificateBuilder in the dev version of 42, and it accepts either naive or non-naive date times, normalizing to naive before it calls into the rust code, so I think we're ok there.

I'm guessing cryptography 42 will probably release relatively soon, possibly sooner than we will, so it's probably not worth merging this branch.  I'll redo things when 42 comes out.

jlaine:
cryptography 42 is out :)

rthalley:
I'm closing this PR in favor of a new one for cryptography 42 on a different branch.

PR #441, opened 2024-01-07
Explicitly track whether FIN has been acknowledged 

jlaine:
A stream's sender part is only finished if all the data and the FIN have been acknowledged. Tracking the acknowledged data ranges is not sufficient to know whether the FIN has been acknowledged too, as the FIN may have been sent on its own. To fix this we need to explicitly keep track of whether the frame containing the FIN was acknowledged.
    
We also:
    
- Remove a bogus assignment of `QuicStreamSender.send_buffer_empty`, this should have been `QuicStreamSender.buffer_is_empty`.
- Only set `buffer_is_empty` to `False` when data or a FIN have been lost. This saves useless calls to `QuicStreamSender.get_frame()`.

This is an alternative to #439

### DIFF
--- a/src/aioquic/quic/connection.py
+++ b/src/aioquic/quic/connection.py
@@ -3019,7 +3019,7 @@ class QuicConnection:
                 QuicFrameType.CRYPTO,
                 capacity=frame_overhead,
                 handler=stream.sender.on_data_delivery,
-                handler_args=(frame.offset, frame.offset + len(frame.data)),
+                handler_args=(frame.offset, frame.offset + len(frame.data), False),
             )
             buf.push_uint_var(frame.offset)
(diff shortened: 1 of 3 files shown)


jlaine:
@mhils Could you confirm this fixes the issue you encountered?

I have naturally added you as a co-author!

mhils:
This looks fantastic, thanks so much! 😃 🍰

jlaine:
> This looks fantastic, thanks so much! 😃 🍰

Cool! This is available from PyPI in version 0.9.25 :smiley:

jlaine:
Oh also, this made me dig a little deeper and I found a potential issue related to RESET, fixed in #444

PR #452, opened 2024-01-14
Adapt "no subjectaltname" test for service-identitity >= 24

jlaine:
When a certificate contains no subjectAltName extension, `service-identity` now raises a `CertificateError` instead of a `VerificationError`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ dependencies = [
     "cryptography",
     "pylsqpack>=0.3.3,<0.4.0",
     "pyopenssl>=22",
-    "service-identity>=23.1.0",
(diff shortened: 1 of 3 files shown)


jlaine:
@rthalley I'm not entirely comfortable with this code, because `extract_patterns` could also raise an Exception..

rthalley:
You could also go for "belt and braces" and just fall back to the first exception if `extract_patterns` raises, e.g.

```
try:
    patterns = service_identity.cryptography.extract_patterns(certificate)
except Exception:
    patterns = []
if len(patterns) == 0:
     errmsg = str(exc)
```

jlaine:
I'm going to merge as-is after thinking about it, we can always revisit this.

PR #454, opened 2024-01-14
Improve documentation for `QuicConnectionProtocol.transmit()`

jlaine:


jlaine:
@ravenblackx This one is for you

ravenblackx (approved):
It's an improvement, but it would be nice to cross-reference `transmit` from the functions that are likely to make calling it necessary (`send_data`, `send_headers`, maybe the datagram one). As someone trying to use the library, I didn't even notice that `transmit` existed, so I wouldn't have seen documentation telling me to call it even with this update. It's surprising when a function named `send` does not actually send, so making that clear in the docs would be helpful. :)

jlaine:
Hm, in that case it's going to mean taking a different approach. Currently, (nearly) all the methods that queue data mention you should call the `QuicConnection.datagrams_to_send` method. This is exactly what `transmit()` does for you if you use the `asyncio` layer. 

 I am not comfortable sprinkling references to the `asyncio` layer all over the place as you can absolutely use `quic` and `h3` and NOT use `asyncio`.

jlaine:
@ravenblackx take a look at the result here for instance:

https://aioquic.readthedocs.io/en/latest/h3.html#aioquic.h3.connection.H3Connection.send_headers

ravenblackx:
Yep, that's a lot more helpful.

PR #455, opened 2024-01-15
ec.generate_private_key() is supposed to be given a curve instance.

rthalley:
In this one place we gave a class instead, which Cryptography < 42.0.0 tolerates and fixes, but it will be deprecated in 42.0.0.


jlaine:
Thanks, go for it :)

PR #461, opened 2024-01-23
Use timezone aware UTC time starting with Cryptography 42.0.0.

rthalley:
This updates cryptography to >= 42.0.0 and makes all needed changes for certificate datetimes to be timezone aware vs. naive.  This fixes deprecation warnings on utcnow() that started with Python 3.12.0.  We also update pyopenssl to >= 24 as version 23 had a cryptography < 42 restriction.


rthalley:
Note that the lint actually passes for me, and the mypy lint errors are a false positive due to importing types-cryptography which isn't needed any more.  This was 1) obscuring actual (unimportant!) type errors which I fixed in my patch, and 2) generating false positive type errors due to not knowing about all the UTC changes.  The patch fixes the GitHub workflow too, but it's not running the fixed workflow from the patch because it is not merged.

PR #467, opened 2024-02-06
Ignore non-ASCII ALPN [#374].

rthalley:
Much like PR #375, this ignores non-ASCII APLNs, but unlike #375 there is an explicit protocol for skipping things when pulling a list.  The ALPN decoder catches decoding exceptions and uses the protocol to say "ignore this list element".  There
is also a test.

We could probably accept anything that decoded as UTF-8 instead of just ASCII as I think the mapping from UTF-8 to Unicode code points is an isomorphism, but I kept the ASCII behavior as I'm really only trying to fix the greasing issue.

Note that there are a few changes not related to this fix caused by black formatting tls.py.

### DIFF
--- a/src/aioquic/tls.py
+++ b/src/aioquic/tls.py
@@ -407,15 +407,25 @@ def push_block(buf: Buffer, capacity: int) -> Generator:
 # LISTS
 
 
+class SkipItem(Exception):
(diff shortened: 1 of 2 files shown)


jlaine (approved):
Looks good, I like the `SkipItem` exception!

Issue #465, opened 2024-01-30
HTTP/3 headers need further validation


lpisu98:
I saw [Issue 21](https://github.com/aiortc/aioquic/issues/21) about header validation, and I want to suggest additional validations. Otherwise, request smuggling vulnerabilities can arise. 
The suggestions are all based on the HTTP/3 RFC specification, which refers to the HTTP/2 RFC specification (See [RFC 9113 - Field Validity](https://datatracker.ietf.org/doc/html/rfc9113#name-field-validity))

- Content-Length should be validated against the actual size of the body
- Transfer-Encoding headers can only be accepted if the value is "trailers"
- Other than rejecting uppercase letters (this is already validated), field names must not contain characters in the ranges 0x00-0x20, 0x41-0x5a, or 0x7f-0xff
- Except for pseudo-headers, headers must not contain single colons (:)
- Header values must not contain null bytes, \n (newlines) and \r (carriage returns)
- Header values must not contain \t (tabs) or spaces at the start or end

PR #472, opened 2024-02-19
More H3 header validation [#465].

rthalley:
This validates that:

* Content-Length should be validated against the actual size of the body

* Transfer-Encoding headers can only be accepted if the value is "trailers"

* Field names must not contain characters in the ranges 0x00-0x20, 0x41-0x5a, or 0x7f-0xff

* Except for pseudo-headers, headers must not contain single colons (:)

* Header values must not contain null bytes, \n (newlines) and \r (carriage returns)

* Header values must not contain \t (tabs) or spaces at the start or end

### DIFF
--- a/src/aioquic/h3/connection.py
+++ b/src/aioquic/h3/connection.py
@@ -25,6 +25,13 @@ logger = logging.getLogger("http3")
 H3_ALPN = ["h3", "h3-32", "h3-31", "h3-30", "h3-29"]
 RESERVED_SETTINGS = (0x0, 0x2, 0x3, 0x4, 0x5)
 UPPERCASE = re.compile(b"[A-Z]")
+COLON = 0x3A
+NUL = 0x00
+LF = 0x0A
+CR = 0x0D
+SP = 0x20
+HTAB = 0x09
+WHITESPACE = (SP, HTAB)
 
 
 class ErrorCode(IntEnum):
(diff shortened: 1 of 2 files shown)


rthalley:
Most of the checking was easy, but the content-length was a bit intrusive due to needing to track what we've seen, and the fact that between optimizations and the fact that streams can end in many ways, the required checks have to be done in several places.  Test coverage is still 100%.

src/aioquic/h3/connection.py
```diff
@@ -194,19 +200,46 @@ def stream_is_request_response(stream_id: int):
     return stream_id % 4 == 0
 
 
+def validate_key(key: bytes):
```
jlaine, line 203: validate_header_name might be a better name?

Also maybe explicitly type the return value to None and a docstring?

src/aioquic/h3/connection.py
```diff
@@ -194,19 +200,46 @@ def stream_is_request_response(stream_id: int):
...
+
+
+def validate_value(key: bytes, value: bytes):
+    # RFC 9113 section 8.2.1
+    for c in value:
+        if c == NUL or c == LF or c == CR:
+            raise MessageError("Header %r value has forbidden characters" % key)
+    if len(value) > 0:
+        first = value[0]
+        if first == SP or first == TAB:
```
jlaine, line 221: Are CR or LF whitespace too?

src/aioquic/h3/connection.py
rthalley: can do

src/aioquic/h3/connection.py
rthalley: CR and LF (along with NUL) are forbidden anywhere in the value by the RFC whereas SP and TAB are only banned as initial or final characters.  I contemplated making WHITESPACE(x) == (x == SP or x == TAB) but didn't as I in other situations WHITESPACE often includes CR and LF, and I wanted to make the translation from the RFC as literal as possible.  But I'm happy with making a WHITESPACE if you prefer it.

src/aioquic/h3/connection.py
```diff
@@ -194,19 +200,46 @@ def stream_is_request_response(stream_id: int):
...
+    for i, c in enumerate(key):
+        if c <= 0x20 or (c >= 0x41 and c <= 0x5A) or c >= 0x7F:
+            raise MessageError("Header %r contains invalid characters" % key)
+        if c == COLON and i != 0:
+            # Colon not at start, definitely bad.  Keys starting with a colon
+            # will be checked in pseudo-header validation code.
+            raise MessageError("Header %r contains a non-initial colon" % key)
+
+
+def validate_value(key: bytes, value: bytes):
```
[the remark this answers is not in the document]
rthalley: ok

src/aioquic/h3/connection.py
```diff
@@ -230,6 +263,20 @@ def validate_headers(
...
+            if key == b"content-length":
+                try:
+                    content_length = int(value)
+                    if content_length < 0:
+                        raise ValueError
+                except ValueError:
+                    raise MessageError("content-length is not a non-negative integer")
+                if stream:
+                    stream.expected_content_length = content_length
+            elif key == b"transfer-encoding" and value != "trailers":
```
jlaine, line 276: Shouldn't this be `b"trailers"` ? I'm not sure we have a test for the happy-path here :)

tests/test_h3.py
```diff
@@ -1701,9 +1830,67 @@ def test_validate_request_headers(self):
...
+        )
+
+        # header with non-initial colon
+        with self.assertRaises(MessageError) as cm:
+            validate_request_headers([(b"x-f:oo", b"foo")])
+        self.assertEqual(
+            cm.exception.reason_phrase, "Header b'x-f:oo' contains a non-initial colon"
         )
 
+        # bad transfer-encoding
```
jlaine, line 1859: We're missing a test with a correct `transfer-encoding` :)

jlaine:
This is great thanks!

src/aioquic/h3/connection.py
jlaine: I'd be fine with `WHITESPACE = (CR, LF) ` for instance, with a comment explaning why we don't need to test other sorts of whitespace.

rthalley:
All feedback incorporated and the trailers bug fixed and tested both ways.  Will merge when the tests complete.

Issue #466, opened 2024-02-04
Idle timeout should be a small value between local idle timeout and remote max idle timeout from transport parameters 


Shouhua:
hi,
I read the doc about [idle timeout](https://datatracker.ietf.org/doc/html/rfc9000#name-idle-timeout) and found that idle timeout should be a small value between local and remote idle timeout.

_`Each endpoint advertises a max_idle_timeout, but the effective value at an endpoint is computed as the minimum of the two advertised values (or the sole advertised value, if only one endpoint advertises a non-zero value)`_

But our code just use the self._configration.idle_timeout: https://github.com/aiortc/aioquic/blob/74e8479e216f53df6d7625596f1b6d4a84477e84/src/aioquic/quic/connection.py#L1060
So is there anything I haven't considered about the code?

rthalley:
Fixed, thanks for your report!

PR #474, opened 2024-02-25
Set the idle timer according to RFC 9000 section 10.1 [#466].

rthalley:
This applies the rules for determining the idle timeout in RFC 9000 section 10.1.

If the remote has note specified a max idle timeout, we use the configured local value as the basis, otherwise we use the smaller of the two.  Given the basis timeout, we then pick the bigger of it and 3 * PTO, to avoid the value being excessively small.

### DIFF
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -2196,6 +2196,48 @@ class QuicConnectionTest(TestCase):
         )
         client._parse_transport_parameters(data)
 
+    def test_parse_transport_parameters_idle_time(self):
+        # Remote idle of 10s and local idle of 60s.
+        client = create_standalone_client(self)
+        data = encode_transport_parameters(
+            # Note the timeout parameter here is in milliseconds.
+            QuicTransportParameters(
+                original_destination_connection_id=client.original_destination_connection_id,
+                max_idle_timeout=10000,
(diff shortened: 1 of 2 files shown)


jlaine:
Looks good, but I think I'd prefer this to be a private `_idle_timeout()` method?

tests/test_connection.py
```diff
@@ -2196,6 +2196,45 @@ def test_parse_transport_parameters(self):
...
+            # Note the timeout parameter here is in milliseconds.
+            QuicTransportParameters(
+                original_destination_connection_id=client.original_destination_connection_id,
+                max_idle_timeout=10000,
+            )
+        )
+        client._parse_transport_parameters(data)
+        self.assertAlmostEqual(client._remote_max_idle_timeout, 10.0)
+        self.assertAlmostEqual(client.idle_timeout(), 10.0)
+        # remote 120s > local 60s
```
jlaine, line 2212: Sole blank lines between the 3 scenarios would help readability.

Also: I know I've haven't done been good at doing this in the past, but I think all comments going forward should start with a capital and end with a dot :)

rthalley:
feedback incorporated, will merge after tests pass

Issue #125, opened 2020-07-20
Only one active stream even using multiple streams | Head of Line Blocking


qyynuaa:
I'm trying to download multiple resources with multiple streams using http3_client.py and http3_server.py. The purpose is to understand the benefit of avoiding Head of Line Blocking using QUIC.

I run the HTTP3 server:
```
python3 examples/http3_server.py --certificate tests/ssl_cert.pem --private-key tests/ssl_key.pem
```
I run the example client to perform four HTTP/3 requests:
``` 
python examples/http3_client.py --ca-certs tests/pycacert.pem --quic-log http3 https://localhost:4433/100000 https://localhost:4433/100000 https://localhost:4433/100000 https://localhost:4433/100000 
```

The result shows that the HTTP requests from the client are sent to the server in parallel,  but the server will respond with only one active stream at any time. 

![image](https://user-images.githubusercontent.com/3302968/87990661-77231a00-cab2-11ea-9659-3cb2b3108e17.png)

I think based on QUIC protocol, we can have multiple active streams at the same time, right? I also find an public example from https://qvis.edm.uhasselt.be/#/files  -->  Option 4 Load a massive demo file. 

![image](https://user-images.githubusercontent.com/3302968/87991137-848cd400-cab3-11ea-95b9-3fb43b3193be.png)

I double checked the source code and I found the current implementation is directly writing all the response body to a byte array, it seems keeps sending the content in this bytearray until it is empty and then switch to start sending another stream.

Did I miss something or are there some configuration parameters I should look at?  

Thanks!

jlaine:
You are right, the current send loop will drain a stream as much as possible before moving on to the next one.

I had once written a naive patch which shuffled the streams here:

https://github.com/aiortc/aioquic/blob/d381ef9825b1d36ae1884a0b43452f89bbb67ad7/src/aioquic/quic/connection.py#L2514

.. however performance took a significant hit. You are welcome to suggest an alternative implementation!

dimitsqx:
What I see in the current implementation is when you create a stream from the client-side by calling:
https://github.com/aiortc/aioquic/blob/c99b43f4c7a1903e9ab2431932395bb7e0b29232/src/aioquic/asyncio/protocol.py#L63-L75

You get the stream_id and create a reader, writer object, however, you don't create a new entry in QuicConnection._streams. When a new stream is created it gets the same stream_id of 0 because there are no entries in QuicConnection._streams. I believe the create_stream should also call: 
[...]

PR #475, opened 2024-02-26
Fairer stream write scheduling [#125].

rthalley:
This greatly improves multiplexing and stream fairness in the use case from #125 with no significant cost (the run I'll show is slightly slower, but I think that could be measurement noise and if not it is small).  The idea is that we keep an ordering for the streams and every time we generate datagrams to send, we recompute the list, moving anyone who got stream data transmitted to the end.

All the tests pass for me.  There seems to be an intermittent loss of one line of coverage related to qlog that I don't understand; we'll see if it shows up in the CI.

Here the is before case, in the terminal and qvis:

Screenshot 2024-02-25 at 17 33 33

Screenshot 2024-02-25 at 17 34 03

And now the much nicer after case:

Screenshot 2024-02-25 at 17 31 29

Screenshot 2024-02-25 at 17 32 09

### DIFF
--- a/src/aioquic/quic/connection.py
+++ b/src/aioquic/quic/connection.py
@@ -359,6 +359,7 @@ class QuicConnection:
         self._spin_highest_pn = 0
         self._state = QuicConnectionState.FIRSTFLIGHT
         self._streams: Dict[int, QuicStream] = {}
+        self._streams_queue: List[QuicStream] = []
         self._streams_blocked_bidi: List[QuicStream] = []
         self._streams_blocked_uni: List[QuicStream] = []
         self._streams_finished: Set[int] = set()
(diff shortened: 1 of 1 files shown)


jlaine:
Thanks so much for tackling this, it's long overdue!

I'm going to have to re-read this, as I'm not sure what happens in certain situations. For instance if you have a mix of streams being discarded (finished) and another stream which hits an exception.. won't we end up keeping the discarded stream in the queue?

rthalley:
I fixed the exception handling.  I also tested again, with bigger transfers and with qlog turned off.  The new code is actually a bit faster, so I'm not worried about performance any more.

jlaine (approved):
This looks great, thanks so much!

PR #299, opened 2022-06-14
add a protocol and is_closing to fix drain on writers

jmgurney:
This fixes issue #250 

This fix can be demonstrated by the tests that will be shortly committed to ntunnel: https://www.funkthat.com/gitea/jmg/ntunnel .

PR #477, opened 2024-02-26
Fix stream writer draining exceptions.

rthalley:
This fix is the code submitted by jmgurney in [#299] with one line added to the tests to give 100% coverage.


rthalley:
I realize this API is under consideration for removal, though I think there are good reasons to have a blocking stream API as I've rolled my own multiple times for all my own use cases (I'm not doing asyncio, or if I am I want a unified API so I haven't been using this API myself.  I'm also happy to see this fixed while we are pondering as it reduces the spam from the test suite.

jlaine:
I know I've been threatening to pull this code for a while, but since the code is still around, might as well do it right.

jlaine (approved):

jlaine:
I think it would be worth adding @jmgurney in a `Co-authored-by`  line here, it is 90% his patch :)

Issue #476, opened 2024-02-26
Specifying cafile loads the CA file in addition to certifi and not instead of certifi


rthalley:
In [dnspython issue 1061](https://github.com/rthalley/dnspython/issues/1061) we were surprised that passing a cafile for the TLS configuration loads the CA in addition to the certifi set instead of replacing it.  My impression is that other TLS software I've used replaces.

Do you think this is a bug?

rthalley:
I tested TLS behavior with httpx and python's native TLS via dnspython's dns.query.https() and dns.query.tls() methods, and they both raise certificate validation failures if I specify a cafile that has just a local CA root.  So, at least for OpenSSL which these will both be using, the behavior is replace not augment.

tykling:
I would absolutely also say +1 for replacing, not augmenting, when a custom CA path is specified. It would match every implementation I've come across. This could also be fixed with a new option to disable loading the system CAs.

But usually it is left up to the system administrator to maintain a combined CA store (system CAs + custom CA) if one is needed.

rthalley:
Fixed, thanks!

PR #479, opened 2024-03-02
Only load certifi if no alternates have been specified. [#476]

rthalley:
This makes tls.py only load the certifi CA info if no alternate CA configuration has been specified.

### DIFF
--- a/src/aioquic/tls.py
+++ b/src/aioquic/tls.py
@@ -271,7 +271,11 @@ def verify_certificate(
 
     # load CAs
     store = crypto.X509Store()
-    store.load_locations(certifi.where())
+
+    if cadata is None and cafile is None and capath is None:
+        # Load defaults from certifi.
+        store.load_locations(certifi.where())
+
     if cadata is not None:
         for cert in load_pem_x509_certificates(cadata):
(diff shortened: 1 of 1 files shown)


rthalley:
I have tested this by hand and it still has 100% coverage, but there is no unit test specifically that a certificate valid from certifi's POV is NOT valid if a cafile with just a self-signed cert is provided.  I haven't come up with a good way to do this yet, as I didn't want to check a real but probably short-lived certificate into the repo, nor did I want to require live Internet.

jlaine:
> I have tested this by hand and it still has 100% coverage, but there is no unit test specifically that a certificate valid from certifi's POV is NOT valid if a cafile with just a self-signed cert is provided. I haven't come up with a good way to do this yet, as I didn't want to check a real but probably short-lived certificate into the repo, nor did I want to require live Internet.

Yeah I don't see a good way of testing this, but I'll live with it.

jlaine (approved):

jlaine:
Uhoh, CI failed :

https://github.com/aiortc/aioquic/actions/runs/8225502073/job/22490610307

PR #481, opened 2024-03-08
Comply with RFC 9000 section 14.1

rthalley:
We were accepting some initial packets that were too small [#401].

We were also emitting datagrams that were too small for client initial acks and server initial ack-eliciting packets.

### DIFF
--- a/src/aioquic/quic/connection.py
+++ b/src/aioquic/quic/connection.py
@@ -785,6 +785,24 @@ class QuicConnection:
                     )
                 return
 
+            # RFC 9000 section 14.1 requires servers to drop all initial packets
+            # contained in a datagram smaller than 1200 bytes.
+            if (
+                not self._is_client
+                and header.packet_type == PACKET_TYPE_INITIAL
+                and len(data) < SMALLEST_MAX_DATAGRAM_SIZE
+            ):
+                if self._quic_logger is not None:
+                    self._quic_logger.log_event(
+                        category="transport",
+                        event="packet_dropped",
+                        data={
+                            "trigger": "initial_packet_datagram_too_small",
+                            "raw": {"length": buf.capacity - start_off},
+                        },
+                    )
+                return
+
             # check destination CID matches
             destination_cid_seq: Optional[int] = None
             for connection_id in self._host_cids:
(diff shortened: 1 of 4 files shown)


rthalley:
We don't currently have super complicated coalescing, and when we detect a need for padding we just do it.  This may result in some lost coalescing opportunities, but I don't want to attempt anything that complicated at the moment.

A number of tests had to be updated due to datagram counts changing.  I recomputed some of the numbers by hand to make sure they were correct, but after that I just updated the remaining tests only checking for sanity.  I.e. if the new values I got looked reasonable, I updated the test to accept them.  In cases where additional datagrams were emitted, I transmitted those values to the other peer in the test.

I wrote an additional test to purposely send too-short data to the server, and verify that the server follows 14.1 and drops it.  I check this by recording a qlog and making sure the drop event is in it.

I haven't seen any problems interoperating with other servers after making this fix, and we continue to interoperate with old versions of ourselves as the invalid short packets we now drop are acks, and things seem to catch up as the handshake succeeds.

jlaine:
Wow, this is interesting. I was aware we were not rejecting some INITIAL which we should have rejected but not that we were emitting insufficiently padded packets.  I think this should be two different PRs or at the very least two different commits.

For the part where a server rejects insufficiently padded INITIAL packets : I see the RFC allows us to close the connection with a protocol violation, any reason for not doing this?

rthalley:
I can split this into two PRs.   I agree the RFC allows us to close immediately, and I didn't do it just because I was trying to have a light touch on the tests.  I can try to do the immediate close when I split it.  That will probably happen on Monday too.

jlaine:
> I can split this into two PRs. I agree the RFC allows us to close immediately, and I didn't do it just because I was trying to have a light touch on the tests. I can try to do the immediate close when I split it. That will probably happen on Monday too.

I think I would rather close the connection, because I don't see how we can sanely recover from the situation. If the remote peer isn't padding their packets sufficiently:

- Either it's a bug in their implementation and I don't see why re-transmissions would be implemented differently.
- Or it's malicious, in which case we don't really want to wait for the connection to timeout, or be held up indefinitely.

jlaine (approved):

PR #189, opened 2021-05-04
[connection] update path challenge according to the draft [JF & CC]

ElNiak:
> " Path validation succeeds when a PATH_RESPONSE frame is received that
>    contains the data that was sent in a previous PATH_CHALLENGE frame.
>    A PATH_RESPONSE frame received on any network path validates the path
>    on which the PATH_CHALLENGE was sent."

We try to modify the PATH_CHALLENGE part of the connection.py to follow this rule that we found with a testing tool Ivy. The problem before was that if the PATH_RESPONSE was received on a different path the challenge was considered as failed. 

JF & CC

PR #483, opened 2024-03-10
Improved path challenge handling.

rthalley:
1) We are willing to reply to up to 5 outstanding challenges on a connection, as it is legal for a client to send more than one.  We limit to 5 to prevent excessive challenging.

2) We now accept path challege responses on any network path, in accordance with RFC 9000 section 8.2.3.

### DIFF
--- a/src/aioquic/quic/connection.py
+++ b/src/aioquic/quic/connection.py
@@ -75,6 +75,8 @@ EPOCH_SHORTCUTS = {
     "1": tls.Epoch.ONE_RTT,
 }
 MAX_EARLY_DATA = 0xFFFFFFFF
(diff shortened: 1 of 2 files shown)


rthalley:
This patch addresses the issue raised by [#189], though with a different implementation.  It also addresses another issue, which is that we may need to reply to more than one remote challenge on the same path.  Both sides are bounded, currently to 5 items, to avoid any possible resource attack.  There are tests and coverage remains at 100%.

rthalley:
I will lint this one further and push an update tonight or tomorrow morning

jlaine (approved):
Looks great, thanks for the explicit comments!

PR #484, opened 2024-03-11
Update GitHub actions to their latest version

jlaine:
This should fix deprecation warnings about Node.js 16 actions.


jlaine:
I don't understand why the coverage has changed?!

rthalley:
> I don't understand why the coverage has changed?!

I think codecov is confused, as when you click through to see what has change it tells you there is 100% coverage!

Issue #501, opened 2024-05-27
[SECURITY] Accepting and storing an unlimited number of CRYPTO frames within a single connection


k4ra5u:
- Aioquic may infinitely receive `CRYPTO` frames within a single connection, rapidly depleting memory and subsequently being forcefully closed by the operating system, leading to a denial of service attack.
- In line 1613 of `quic/connection.py`, the server only checks `offset + length < 2^62 - 1` when processing `CRYPTO` frames, and then stores their contents in `QuicConnection._crypto_streams[Epoch.ONE_RTT]` , resulting in memory consumption.
- To validate the effect, I simulated an attacker sending `CRYPTO` frames with an Offset set to 0x1000, but with a length of only 0x200 to prevent some memory merging operations. As shown in the graph, within the 90s of the attack occurrence, aioquic consumed 100GB of memory and will be killed soon by the operating system.
![1](https://github.com/aiortc/aioquic/assets/35438062/0402fccb-3c58-40a9-a279-7d8fc3b22923)

rthalley:
This should be fixed now, thanks for your report.

PR #505, opened 2024-06-15
Only buffer up to 512 KiB of pending CRYPTO frames [#501].

rthalley:
Address #501 by limiting the size required to buffer pending CRYPTO frames to 512 KiB.  I picked this size based on it being big enough for post-quantum crypto, but still not very big.  The limit applies to the amount of storage we need to allocate, not the size of inbound fragments, so e.g. a bunch of small fragment around the 512 KiB offset would still trigger as we use a contiguous buffer and the required size would exceed the limit.

### DIFF
--- a/src/aioquic/quic/connection.py
+++ b/src/aioquic/quic/connection.py
@@ -57,11 +57,7 @@ from .packet import (
     push_ack_frame,
     push_quic_transport_parameters,
 )
-from .packet_builder import (
-    QuicDeliveryState,
-    QuicPacketBuilder,
-    QuicPacketBuilderStop,
-)
+from .packet_builder import QuicDeliveryState, QuicPacketBuilder, QuicPacketBuilderStop
 from .recovery import QuicPacketRecovery, QuicPacketSpace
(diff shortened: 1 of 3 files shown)


rthalley:
Note that the codespell check seems confused; it's complaining about a test function name "assertIn" in code I didn't change, and is just confused.  When I run codespell locally, I don't see it.  I see there is some issue with the openssl vendoring too, though again not related to this PR.

src/aioquic/quic/connection.py
```diff
@@ -93,6 +93,7 @@
 STREAM_COUNT_MAX = 0x1000000000000000
 UDP_HEADER_SIZE = 8
 MAX_PENDING_RETIRES = 100
+MAX_PENDING_CRYPTO = 524288
```
jlaine, line 96: Maybe add a comment stating this is a number of bytes?

jlaine:
If you rebase on top of `main` the CI errors should be fixed. Once CI is green feel free to merge, thanks!

jlaine (approved):

Issue #499, opened 2024-05-13
No support for ECC certs with 384 key size, logs doesn't suggest the reason of failure


Szwendacz99:
The library is not able to handle endpoints using ECC TLS certificate with 384bit key size. Example is `quic.nginx.org` domain.

It also fails in a way that is not easy to debug and find the reason:

```text
❯ python /tmp/http3_client.py -v https://quic.nginx.org
2024-05-13 16:35:27,075 DEBUG asyncio Using selector: EpollSelector
2024-05-13 16:35:27,084 DEBUG quic [c77d98a740b89af2] TLS State.CLIENT_HANDSHAKE_START -> State.CLIENT_EXPECT_SERVER_HELLO
2024-05-13 16:35:27,114 INFO quic [c77d98a740b89af2] Retrying with token (66 bytes)
2024-05-13 16:35:27,116 DEBUG quic [c77d98a740b89af2] TLS State.CLIENT_HANDSHAKE_START -> State.CLIENT_EXPECT_SERVER_HELLO
2024-05-13 16:35:27,145 DEBUG quic [c77d98a740b89af2] QuicConnectionState.FIRSTFLIGHT -> QuicConnectionState.CONNECTED
2024-05-13 16:35:27,146 INFO quic [c77d98a740b89af2] Connection close received (code 0x128, reason handshake failed)
2024-05-13 16:35:27,146 DEBUG quic [c77d98a740b89af2] QuicConnectionState.CONNECTED -> QuicConnectionState.DRAINING
2024-05-13 16:35:27,746 DEBUG quic [c77d98a740b89af2] Discarding epoch Epoch.INITIAL
2024-05-13 16:35:27,747 DEBUG quic [c77d98a740b89af2] Discarding epoch Epoch.HANDSHAKE
2024-05-13 16:35:27,747 DEBUG quic [c77d98a740b89af2] Discarding epoch Epoch.ONE_RTT
2024-05-13 16:35:27,747 DEBUG quic [c77d98a740b89af2] QuicConnectionState.DRAINING -> QuicConnectionState.TERMINATED
Traceback (most recent call last):
  File "/tmp/http3_client.py", line 564, in <module>
    asyncio.run(
  File "/usr/lib64/python3.12/asyncio/runners.py", line 194, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
[...]

PR #509, opened 2024-06-18
Enable SHA-384 based signature algorithms

jlaine:
The signature algorithms we offer match those offered by Chromium 126.

Fixes: #499


rthalley (approved):

jlaine:
@rthalley I've pushed a second commit to support SECP384R1 key exchange, still OK with you?

rthalley:
Still ok w/ it!

Issue #513, opened 2024-06-19
QLOG not working in HTTP/3 server demo?


lminiero:
Hi,

first of all, let me thank you for your code: it's been invaluable so far as a way for me to test QUIC, as I'm currently studying and prototyping it. The QLOG support in particular has been quite helpful when trying to figure out when something wasn't working. That said, it looks like no QLOG seems to be generated, for me, when I enable it in the `http3_server.py` demo: I'm not sure what could be the cause, since it seems to be working fine when I enable it in the DOQ demos, and the HTTP/3 client demo as well. I can't see any obvious reason for this when looking at the code, as it looks like it's enabled the same way in all demos, so it may be something deeper in the stack.

Do you have an idea for what could be the reason?

Thanks!

rthalley:
If I run it with "-q some_directory" I see qlogs in that directory, so I can't reproduce what you are seeing.  Are you sure the server is seeing requests?   What version of aioquic are you using?

jlaine:
Hi @lminiero I'm really puzzled by what your are describing, especially since the demo server at quic.aiortc.org is spewing (large) amounts of QLOG files. Be aware that the files only get written once a connection is closed.

Could you try adding some `print()` statements in `QuicFileLogger` to check it's actually being instantiated and its `end_trace` method is being called?

lminiero:
> Be aware that the files only get written once a connection is closed

Ah this could be the problem I'm encountering, I suspect. I was using this [Webtransport demo page](https://webtransport.day/) with `http3_server.py`, which works, but there's no way to close the connection from the page: even closing or refreshing the page doesn't seem to send a CC. CTRL-Cing `http3_server.py` to close it cleanly apparently is not acting as a trigger to close the connection either, as in my case the folder stays empty.

lminiero:
> Could you try adding some `print()` statements in `QuicFileLogger` to check it's actually being instantiated and its `end_trace` method is being called?

I can confirm that it's being instantiated, but `end_trace` is not called in the scenario I described in my previous comment.

jlaine:
`end_trace` should be called when a QUIC connection gets closed:

https://github.com/aiortc/aioquic/blob/ff3281fa33ca14943e7954b8baf51c1e950c791d/src/aioquic/quic/connection.py#L1275

I just tried the following:

```shell
# Run the server
mkdir qlog
python examples/http3_server.py --certificate tests/ssl_cert.pem --private-key tests/ssl_key.pem --quic-log qlog

[...]

PR #515, opened 2024-06-22
Make `serve()` an async context manager

jlaine:
In order to get a clean shutdown of the asyncio-based QUIC server, we need to:

- Make the `QuicServer.close()` method async.
- Make the `serve()` method an async context manager.

See #513


jlaine:
@rthalley This is a breaking change, so it looks we would need to make the next release 2.0.

rthalley:
I think clean closing is worth the breaking change.

HarshArora-1205:
Status?

PR #529, opened 2024-06-30
Improve connection establishment tests

jlaine:
Before we attempt to refactor packet / datagram padding, we need to finely capture the current behaviour, so:

- Add a detailed test for connection establishment without packet loss.

- Exhaustively check the events emitted by client / server at each step with an `assertEvents` helper.

- Check the unacknowledged sent packets at each step with an `assertSentPackets` helper.

- Make use of the `client_and_server()` helper instead  of manually spinning up client and server instances.


rthalley (approved):
I like it!

PR #532, opened 2024-07-01
Improve padding of coalesced datagrams containing INITIAL

jlaine:
Our previous padding algorithm padded all client-sent or ack-eliciting INITIAL packets to a full datagram size. While this satisfies the specification requirements, the downside is that it made it impossible to coalesce any other packets after the INITIAL.

We now mostly defer the padding decision until the datagram is finalised and perform padding by appending zeroes at the end of the datagram. As an exception to this rule, in the presence of short-header packets we insert the padding inside the packet.


jlaine:
I'm puzzled by the flakiness in the `test_send_max_data_blocked_by_cc` test.. why does Windows end up with an ever-so-slightly different congestion window?

PR #533, opened 2024-07-01
better backwards compatibility with older versions of cryptography

totaam:
Older versions don't have `not_valid_before_utc` or `not_valid_after_utc`.

### DIFF
--- a/src/aioquic/tls.py
+++ b/src/aioquic/tls.py
@@ -227,9 +227,11 @@ def verify_certificate(
 ) -> None:
     # verify dates
     now = utcnow()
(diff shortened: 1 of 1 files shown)


jlaine:
Hi @totaam, we're aware that older `cryptography` versions do not have these attributes, which is why the minimum cryptography version was raised to 42.0.0 in 488f1ab60b36328e5f5565e5d2b4969a6dd6263b.

totaam:
OK, feel free to ignore it.

I'm just leaving this out there as some downstream may need this patch. (I needed it on Fedora 40 which just came out a few months ago)

PR #537, opened 2024-07-08
Support Libressl

catap:
Here a small changes which allows to build and run aioquick against LibreSSL

### DIFF
--- a/tests/test_asyncio.py
+++ b/tests/test_asyncio.py
@@ -196,6 +196,7 @@ class HighLevelTest(TestCase):
             )
         )
 
+    @skipIf("ed448" in SKIP_TESTS, "Skipping ed448 tests")
     @asynctest
     async def test_connect_and_serve_with_ed448_certificate(self):
         await self._test_connect_and_serve_with_certificate(
--- a/tests/test_tls.py
+++ b/tests/test_tls.py
@@ -2,7 +2,7 @@ import binascii
 import datetime
 import ssl
 from functools import partial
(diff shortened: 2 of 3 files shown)


jlaine:
FYI we are considering dropping the C code here in PR #457

catap:
@jlaine well, I've used this patch to make OpenBSD port anyway and I think that better to backport to upstream patches.

tests/test_asyncio.py
```diff
@@ -196,6 +196,7 @@ async def test_connect_and_serve_with_ed25519_certificate(self):
             )
         )
 
+    @skipIf("ed448" in SKIP_TESTS, "Skipping ed448 tests")
```
jlaine, line 199: Where is this used?

tests/test_tls.py
```diff
@@ -584,9 +584,12 @@ def test_handshake_with_ed25519_certificate(self):
         )
 
     def test_handshake_with_ed448_certificate(self):
```
jlaine, line 586: Why not use the same `skipIf("ed448" in SKIP_TESTS)` as above?

tests/test_asyncio.py
catap: I think I don't understand your question. Can I ask you to rephase it?

tests/test_tls.py
catap: Indeed it make things cleaner

catap:
@jlaine and ping here as well

PR #538, opened 2024-07-08
Decrease minimal Hello and Window to tolerate OpenBSD values

catap:


catap:
@jlaine ping here

catap:
@jlaine and ping here

jlaine:
Thanks, sorry for the slow turnaround!

PR #539, opened 2024-08-08
Enable jump to source in docs

jgarte:
Hi,

This enables jumping to the source code in the docs.

wdyt

example from another project of the `[source]` link in the rendered HTML:

image


rthalley:
I use this option in dnspython and think this is a nice thing.

jlaine:
I have no strong opinion here, but if we're going to do it, can we keep the extensions sorted alphabetically please to appease my OCD? :)

jgarte:
Sure, I'll update the PR to sort that.

PR #553, opened 2025-01-15
26 add packet ordering mechanism

nuvious:
Added a packet-reordering mechanism while still preserving high entropy requirements by creating a sorted list of random bytes that are mapped to the chunks of the encrypted payload. These are then shuffled so the prefixes are no longer in sorted order. On receipt, the prefixes are re-mapped to the payload chunks, ordered, and reconstructed.

Additionally the CID was increased to a full 160 bits allowing payloads to be 128 bits. The RSA key exchange was modified to include 4 bytes or random data and to have the chunks of the key increased to 128 bits. This makes all the CIDs a uniform 160 bits and increases the bandwidth as a benefit.


nuvious:
Closing because this was to be merged in my personal project which is a covert channel project. Just opened the pull request on the wrong repo. Meant to PR it into my own fork.

PR #571, opened 2025-04-04
Build wheels for musllinux

jlaine:
We now have binary builds of OpenSSL for `musllinux`, which allows us to build wheels for this platform.


reneleonhardt:
Hello @jlaine can you tag a new release to fix the 7 CVEs since OpenSSL 3.3.0 (used in aioquic 1.2.0 since July)?
https://github.com/openssl/openssl/blob/openssl-3.4.1/CHANGES.md

Thank you for the Alpine support! 😄

PR #580, opened 2025-05-06
Build wheel for Windows arm64

### LINKED ISSUE #579: Build wheel for windows arm64
Now that [the required runner is available](https://github.blog/changelog/2025-04-14-windows-arm64-hosted-runners-now-available-in-public-preview/), this should be possible.

saschanaz:
Fixes #579

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -127,6 +127,8 @@ jobs:
             arch: AMD64
           - os: windows-latest
             arch: x86
+          - os: windows-11-arm
+            arch: ARM64
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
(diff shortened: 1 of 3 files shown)


jlaine:
Yeah it's a bit more involved than this, we need:

- Binary builds of OpenSSL for Windows/ARM64 : https://github.com/aiortc/aioquic-openssl/pull/23
- Binary wheels of pylsqpack for Windows/ARM64 : https://github.com/aiortc/pylsqpack/pull/44

saschanaz:
Ah yes, 3221225477. 0xC0000005. So it's not me. https://github.com/pyca/cryptography/issues/12867

saschanaz:
It's crashing with the same error code in aioquic CI: https://github.com/aiortc/aioquic/actions/runs/14966543755/job/42037656065?pr=580

So it's not just me at least...

saschanaz:
(I think MSVC is fixed, so maybe retrying would fix it)

saschanaz:
The crash is indeed gone but somewhere something hits this:

https://github.com/python/cpython/blob/48d0d0dd9733eae4935f2ebd31bef786d8074fc8/Include/Python.h#L51

> C:\Users\runneradmin\AppData\Local\pypa\cibuildwheel\Cache\nuget-cpython\pythonarm64-freethreaded.3.14.0-rc3\tools\include\Python.h(51): fatal error C1189: #error:  "The limited API is not currently supported in the free-threaded build"

Free-threaded?

saschanaz:
As of ff73320 the only failure is windows 3.11 which seems spurious.

PR #590, opened 2025-07-06
http3: Fix stream memory leak in the H3Connection

G1gg1L3s:
Hello! I was testing the library with Hypercorn and noticed that long-lived HTTP3 connections gradually consume server memory, so I decided to investigate.

I found that the [H3Connection._stream](https://github.com/G1gg1L3s/aioquic/blob/218f940467cf25d364890a602b8fc451ca635062/src/aioquic/h3/connection.py#L387C9-L387C21) dict is never cleared of created streams, leading to a memory leak.

This PR addresses the issue. First, I introduced tracking of whether HTTP/3 streams have ended on both sides. Then, I refactored `H3Connection._get_or_create_stream` to be a context manager which automatically removes streams if they are ended on both sides and are not blocked. Finally, I added a test for this case, though it's a bit dirty because it accesses internal H3Connection field, but it works well overall.

Let me know what you think! I may miss something in the H3 state machine, but overall I tried a couple of approaches and the current one seems to play nicely with the way HTTP3 is implemented.

#### Similar issue in QUIC

By the way, there is a small memory leak on the QUIC level, but it's rather a feature and not a bug :) I'm talking about the [QuicConnection._streams_finished](https://github.com/aiortc/aioquic/blob/e2a401a6430255812a705e82ab0d127d46416d91/src/aioquic/quic/connection.py#L373) set, which is populated with finished streams to prevent their reusage. In theory it means that long-living connections with short-living streams will also consume memory, though it's a slow process.

I've seen a similar approach in Rust [quiche](https://github.com/cloudflare/quiche/blob/762c61833f644374bb58da6d69af6426af1400d4/quiche/src/stream/mod.rs#L99), which tracks used stream IDs in a set.

However, I didn't find anything similar in [quic-go](https://github.com/quic-go/quic-go/blob/fd32cf5c69a5941890d55041313a71a2e42a7773/streams_map_incoming.go#L115) or in [quinn](https://github.com/quinn-rs/quinn/blob/64139181e0fcdf337737d32dac077e96391d9d52/quinn-proto/src/connection/streams/state.rs#L964). I haven’t tested or looked into it deeply, but I suspect that they use some mechanism to avoid storing the entire set. So, it may be worth investigating.

### DIFF
--- a/src/aioquic/h3/connection.py
+++ b/src/aioquic/h3/connection.py
@@ -1,5 +1,7 @@
+import contextlib
 import logging
 import re
+from collections.abc import Generator
 from enum import Enum, IntEnum
 from typing import Dict, FrozenSet, List, Optional, Set
 
@@ -345,7 +347,8 @@ class H3Stream:
         self.blocked = False
         self.blocked_frame_size: Optional[int] = None
         self.buffer = b""
-        self.ended = False
+        self.receiving_ended = False
+        self.sending_ended = False
         self.frame_size: Optional[int] = None
         self.frame_type: Optional[int] = None
         self.headers_recv_state: HeadersState = HeadersState.INITIAL
@@ -356,6 +359,14 @@ class H3Stream:
         self.stream_type: Optional[int] = None
         self.expected_content_length: Optional[int] = None
         self.content_length: int = 0
+
+    def is_ended(self) -> bool:
+        return self.sending_ended and self.receiving_ended and not self.blocked
+
+    def finish_sending(self) -> None:
+        if self.sending_ended:
+            raise FrameUnexpected("stream was already ended")
+        self.sending_ended = True
 
 
 class H3Connection:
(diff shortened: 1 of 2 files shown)


G1gg1L3s:
@jlaine, sure! Done :saluting_face:. Hope it's enough, but let me know if it's not

jlaine:
It looks as though CI has bit-rotted so I'm going to fix those failures. Could you please fix the remaining issues:

- One line of code is not covered by tests for the "double stream ended". Either this is a condition which *cannot* happen no matter what we receive (and so would indicate a programming error) => change this to an assertion. If it *can* happen, please add a test for it :)
- A minor typo ("ontrol" instead of "Control")

G1gg1L3s:
Thanks, fixed!

jlaine:
I've resolved the CI issues in #600 so if you rebase on top of main, CI should work.

G1gg1L3s:
@jlaine, done! Is the description ok?

PR #603, opened 2025-10-07
Limit the number of remote path challenges stored per path

### LINKED ISSUE #544: [SECURITY] aioquic may store an unlimited number of remote path challenges
## Summary
aioquic may infinitely receive `PATH_CHALLENGE` frames and store challenge data in `remote_challenges`, resulting in unbounded memory usage. It may lead to a denial-of-service attack.

## Tested Version
Latest git commit (9bc1e43)

## Details
#483 appears to be the cause of this vulnerability. Although MAX_REMOTE_CHALLENGES is defined, the constant is not used. As a result, the `remote_challenges` queue can grow infinitely.

## Suggestion
https://github.com/aiortc/aioquic/blob/9bc1e43d13be3f06339841aca7c8560825053371/src/aioquic/quic/connection.py#L2045
Check the length of `remote_challenges` before appending challenge data.

jlaine:
In order to avoid resource exhaustion, limit the number of remote path challenges we store per path.If the limit is exceeded, we drop any additional path challenge.

Fixes: #544.

### DIFF
--- a/src/aioquic/quic/connection.py
+++ b/src/aioquic/quic/connection.py
@@ -68,7 +68,7 @@ EPOCH_SHORTCUTS = {
     "1": tls.Epoch.ONE_RTT,
 }
 MAX_EARLY_DATA = 0xFFFFFFFF
-MAX_REMOTE_CHALLENGES = 5
+MAX_REMOTE_CHALLENGES = 32
 MAX_LOCAL_CHALLENGES = 5
 SECRETS_LABELS = [
     [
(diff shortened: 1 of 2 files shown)


reneleonhardt:
Wouldn't a `for _ in range(len - max)` be better in case there is more than one challenge exceeding the limit?

jlaine:
> Wouldn't a `for _ in range(len - max)` be better in case there is more than one challenge exceeding the limit?

This is the only place where we append to the list, so at most it can have exceeded the limit by 1 :)

jlaine:
Thinking about it some more I think it's better to tail drop here. If the path challenges are building up, the remote party is doing something weird. Let them send some ACKs to get things moving again.

PR #605, opened 2025-10-08
Generate all test TLS certificates on the fly

### LINKED ISSUE #557: tests fail after 2028-07-07
While working on [reproducible builds](https://reproducible-builds.org/) for [openSUSE](https://en.opensuse.org/openSUSE:Reproducible_Builds), I found that
our `python-aioquic` 1.2.0 package fails tests after 2028-07-07, because tests/ssl_cert.pem expired.

Please mock the current time or use test certs that are valid for 1000y.

Background:
As part of my work on reproducible builds for openSUSE, I check that software still gives identical build results in the future.
The usual offset is +16 years, because that is how long I expect some software will be used in some places.
This showed up failing tests in our package build.
See https://reproducible-builds.org/ for why this matters.

jlaine:
Instead of relying on fixed certificates which will eventually expire, generate them as part of the test suite.

Fixes: #557


bmwiedemann:
Hi. What is missing to get this merged? It looked good in my test.

Issue #596, opened 2025-09-13
Feature Request: Support for RESET_STREAM_AT Frame Extension (for WebTransport Compliance)


lemonsterfy:
### Is your feature request related to a problem? Please describe.

Yes. The `draft-ietf-webtrans-http3-13` (and subsequent versions) of the WebTransport over HTTP/3 specification now mandates support for the `RESET_STREAM_AT` frame, as defined in `draft-ietf-quic-reliable-stream-reset`.

Currently, `aioquic` implements the standard `RESET_STREAM` frame from RFC 9000. This makes it challenging for higher-level protocol libraries built on top of `aioquic`, such as our project `PyWebTransport`, to be fully compliant with the WebTransport standard. The standard requires this frame to ensure that a predictable amount of stream data is delivered before the stream is reset, which is a critical reliability feature.

### Describe the solution you'd like

We would like to request the implementation of the QUIC reliable stream reset mechanism. This would primarily involve:

1.  Support for parsing the `RESET_STREAM_AT` frame (`0x24`).
2.  Support for serializing and sending the `RESET_STREAM_AT` frame.
3.  An extension to the public API, perhaps a new method like `quic.reset_stream_at(stream_id, error_code, final_size)`, to allow application protocols to trigger this frame with a specified final size.
4.  Internal adjustments to the stream state machine to handle the "draining" period, where data is transmitted up to the `final_size` before the stream is fully reset.

Relevant specification: [draft-ietf-quic-reliable-stream-reset](https://www.ietf.org/archive/id/draft-ietf-quic-reliable-stream-reset-07.txt)

### Describe alternatives you've considered

We considered implementing the frame encoding/decoding logic at the application layer (within our WebTransport library). However, this is not feasible due to the strict architectural boundary between the QUIC transport layer (managed by `aioquic`) and the application layer. QUIC frames must be handled deep within the transport's internal state machine, and there are no hooks to safely add support for a new frame type from the outside.

### Additional context

The adoption of WebTransport is growing, and full compliance is becoming increasingly important for interoperability. Adding support for `RESET_STREAM_AT` would significantly enhance `aioquic`'s position as a foundational library for building WebTransport servers and clients in Python.

Our team working on `PyWebTransport` is willing to contribute to this effort. Before we investigate creating a pull request, we wanted to open this issue to gauge the maintainers' interest and discuss the best approach for a potential implementation.

Thank you for your consideration and for maintaining this excellent library.

jlaine:
[...]

PR #614, opened 2025-10-13
Reliable reset

bashi:
https://datatracker.ietf.org/doc/html/draft-ietf-quic-reliable-stream-reset


gsnedders:
To copy the status here, from https://github.com/aiortc/aioquic/issues/596#issuecomment-3395529469:

> I want to complete it if I have time to work on it (currently I don't have enough time for that). To be honest I don't remember the details and the PR might be stale and/or wrong.

So it would seem like before working on this, one would need someone to first investigate what the state of this PR is.

The informal review in https://github.com/aiortc/aioquic/issues/596#issuecomment-3396797518 seemed to suggest it is in a pretty good shape.

PR #634, opened 2026-06-06
Add Post-Quantum key exchange support (ML-KEM / FIPS 203)

nercone-dev:
## Summary

Adds TLS 1.3 key exchange support for ML-KEM (NIST FIPS 203) — both pure and
hybrid groups — and fixes the server-side group selection logic that this
feature exposed.

### New groups

| Group | TLS code point | Type |
|---|---|---|
| `MLKEM768` | 0x0201 | Pure PQC |
| `MLKEM1024` | 0x0202 | Pure PQC |
| `X25519MLKEM768` | 0x11EC | Hybrid (X25519 + ML-KEM 768) |
| `SECP256R1MLKEM768` | 0x11EB | Hybrid (P-256 + ML-KEM 768) |
| `SECP384R1MLKEM1024` | 0x11ED | Hybrid (P-384 + ML-KEM 1024) |

Groups are advertised only when `default_backend().mlkem_supported()` returns
true (requires OpenSSL 3.5+ and `cryptography >= 48.0.0`).

Hybrid key shares concatenate the classical public key followed by the ML-KEM
encapsulation key, per draft-tls-mlkem-iana. The shared secret is
`mlkem_ss || classical_ss`.

### Server-side group selection fix

Previously the server iterated over the client's `key_share` list in the order
the client sent it. With multiple PQC and classical groups now possible, this
meant the server's own configured group preference (`_supported_groups`) was
effectively ignored.

The server now builds a lookup dict from the client's key shares and iterates
`self._supported_groups` instead, selecting the first mutually supported group
in server-priority order — consistent with how OpenSSL respects
`SSL_CTX_set1_groups`.

### Other changes

- `ClientHello` serialization buffer increased from 1 KiB to 16 KiB (ML-KEM 768
  encapsulation key alone is 1184 bytes).
- Test buffers enlarged accordingly; `ClientHello` size assertion relaxed.
- `cryptography` minimum raised from `>=42.0.0` to `>=48.0.0`.

### Tests

- 5 handshake tests (one per new group), skipped automatically when the backend
  lacks ML-KEM support.
- Group capability detection test.
- 2 key-share size assertions.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ classifiers = [
 ]
 dependencies = [
     "certifi",
-    "cryptography>=42.0.0",
+    "cryptography>=48.0.0",
     "pylsqpack>=0.3.3,<0.4.0",
     "pyopenssl>=24",
     "service-identity>=24.1.0",
--- a/src/aioquic/quic/configuration.py
+++ b/src/aioquic/quic/configuration.py
@@ -118,6 +118,7 @@ class QuicConfiguration:
             QuicProtocolVersion.VERSION_2,
         ]
     )
+    ssl_groups: Optional[str] = None
     verify_mode: Optional[int] = None
 
     def load_cert_chain(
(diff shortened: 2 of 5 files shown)


rthalley:
I tested this against google and cloudflare and got X25519MLKEM768 like I expected, and thing still worked against non-PQ-aware sites.

Thanks for doing this work!

PR #648, opened 2026-08-21
Add max_udp_payload_size to transport parameters.

achristensen07:
aioquic does not handle UDP payloads larger than 1500 bytes.  According to RFC 9000, "The default for this parameter is the maximum permitted UDP payload of 65527."  This causes issues with quic implementations that try to send aioquic UDP payloads larger than 1500 bytes.  Use max_udp_payload_size in the transport parameters to tell such implementations to keep the UDP payload size below 1500 bytes.


achristensen07:
Because there's a new transport parameter, the size of the frame inside the intial packet changed, which requires test updates.

